PR #1, opened 2020-03-18
Add suffixing for multi-upstream connections

delthas:
I've chosen `/` over `@` for suffixing because `@` would mess up irc prefix parsing (in cases like `delthas@baze!~delthas@86.238.46.223`).


schema.sql
```diff
@@ -5,6 +5,7 @@ CREATE TABLE User (
 
 CREATE TABLE Network (
 	id INTEGER PRIMARY KEY,
+	name VARCHAR(255) NOT NULL,
```
emersion: Can we allow this to be NULL, and fallback to `addr` in this case?

delthas:
Makes sense. I'll update the PR soon.

delthas:
Updated.

downstream.go
```diff
@@ -154,12 +150,48 @@ func (dc *downstreamConn) upstream() *upstreamConn {
...
 	if uc := dc.upstream(); uc != nil {
 		return uc, name, nil
 	}
 
-	// TODO: extract network name from channel name if dc.upstream == nil
+	network := ""
+	if i := strings.LastIndexAny(name, "/@"); i >= 0 {
+		network = name[i+1:]
+	}
+	if i := strings.IndexAny(name, "/@"); i >= 0 {
```
emersion: I think we want to use `LastIndexAny` in all cases, because we want `#thing/other-thing/network` to be split into channel name `#thing/other-thing` and network name `network`. The logic is different for usernames because we want to allow things like `user@client-id/network` for instance (`client-id` allows to properly send history to multiple clients).

downstream.go
```diff
@@ -154,12 +150,48 @@ func (dc *downstreamConn) upstream() *upstreamConn {
 	return upstream
 }
 
+func (dc *downstreamConn) marshalEntity(uc *upstreamConn, name string) string {
+	for _, r := range name {
+		switch r {
+		case '#', '&', '+', '!':
```
emersion: Can we add a TODO to use the server's ISUPPORT channel prefixes here?

PR #3, opened 2020-03-18
Add create-network bouncer service command

delthas:


service.go
```diff
@@ -85,3 +92,44 @@ func handleServiceHelp(dc *downstreamConn, params []string) error {
...
+	nick := fs.String("nick", "", "")
+
+	if err := fs.Parse(params); err != nil {
+		return err
+	}
+	if *addr == "" {
+		return fmt.Errorf("flag addr is required")
+	}
+
+	if !strings.ContainsRune(*addr, ':') {
```
emersion: This will be automatically done at connect time, no need to store the hardcoded port in the DB.

(Not storing the hardcoded port allows users to refer to a name-less network more easily, e.g. `chat.freenode.net` instead of `chat.freenode.net:6697`)

service.go
```diff
@@ -58,6 +60,11 @@ func init() {
 			desc:   "print help message",
 			handle: handleServiceHelp,
 		},
+		"create-network": {
+			usage:  "<-addr addr> [-name name] [-username username] [-pass pass] [-realname realname] [-nick nick]",
```
emersion: Maybe `-addr <addr>` would be a little clearer

PR #5, opened 2020-03-20
Add WHOIS support

delthas:


upstream.go
```diff
@@ -603,7 +603,101 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+	case irc.RPL_WHOISIDLE:
+		var nick string
+		if err := parseMessageParams(msg, nil, &nick, nil); err != nil {
+			return err
+		}
+
+		uc.forEachDownstream(func(dc *downstreamConn) {
+			nick := dc.marshalNick(uc, nick)
+			params := []string{dc.nick, nick}
+			params = append(params, msg.Params[2:]...)
```
delthas: forwarding all remaining parameters rather than just sending idle time (as specified in rfc 2812), because some servers (eg rizon) send additional parameters, such as signon time.

delthas:
See #6 .

PR #8, opened 2020-03-21
Multiple features (NOTICE, LIST, logs)

delthas:
_Several commits in one PR for easier rebasing. See individual commits for details._

_Not that Github may show some commits in the wrong order (ordered by date rather than by parent)._

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -19,6 +19,7 @@ type Server struct {
 	TLS       *TLS
 	SQLDriver string
 	SQLSource string
+	LogPath   string
 }
 
 func Defaults() *Server {
@@ -70,6 +71,10 @@ func Parse(r io.Reader) (*Server, error) {
 			srv.TLS = tls
 		case "sql":
 			if err := d.parseParams(&srv.SQLDriver, &srv.SQLSource); err != nil {
+				return nil, err
+			}
+		case "log":
+			if err := d.parseParams(&srv.LogPath); err != nil {
 				return nil, err
 			}
 		default:
--- a/downstream.go
+++ b/downstream.go
@@ -1077,6 +1077,45 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 			}
 			sendTopic(dc, ch)
 		}
+	case "LIST":
+		// TODO: support ELIST when supported by all upstreams
+
+		dc.user.pendingLISTsLock.Lock()
+		defer dc.user.pendingLISTsLock.Unlock()
+
+		pl := pendingLIST{
+			downstreamID:    dc.id,
+			pendingCommands: make(map[int64]*irc.Message),
+		}
+		var upstreamChannels map[int64][]string
+		if len(msg.Params) > 0 {
+			upstreamChannels = make(map[int64][]string)
+			channels := strings.Split(msg.Params[0], ",")
+			for _, channel := range channels {
+				uc, upstreamChannel, err := dc.unmarshalEntity(channel)
+				if err != nil {
+					return err
+				}
+				upstreamChannels[uc.network.ID] = append(upstreamChannels[uc.network.ID], upstreamChannel)
+			}
+		}
+
+		dc.user.pendingLISTs = append(dc.user.pendingLISTs, pl)
+		dc.forEachUpstream(func(uc *upstreamConn) {
+			var params []string
+			if upstreamChannels != nil {
+				if channels, ok := upstreamChannels[uc.network.ID]; ok {
+					params = []string{strings.Join(channels, ",")}
+				} else {
+					return
+				}
+			}
+			pl.pendingCommands[uc.network.ID] = &irc.Message{
+				Command: "LIST",
+				Params:  params,
+			}
+			uc.trySendList(dc.id)
+		})
 	case "NAMES":
 		if len(msg.Params) == 0 {
 			dc.SendMessage(&irc.Message{
--- a/downstream.go
+++ b/downstream.go
@@ -1099,7 +1099,7 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 				sendNames(dc, ch)
 			} else {
 				// NAMES on a channel we have not joined, ask upstream
-				uc.SendMessageLabeled(dc, &irc.Message{
+				uc.SendMessageLabeled(dc.id, &irc.Message{
 					Command: "NAMES",
 					Params:  []string{upstreamChannel},
 				})
@@ -1146,7 +1146,7 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 			params = []string{upstreamName}
 		}
 
-		uc.SendMessageLabeled(dc, &irc.Message{
+		uc.SendMessageLabeled(dc.id, &irc.Message{
 			Command: "WHO",
 			Params:  params,
 		})
@@ -1203,7 +1203,7 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 			params = []string{upstreamNick}
 		}
 
-		uc.SendMessageLabeled(dc, &irc.Message{
+		uc.SendMessageLabeled(dc.id, &irc.Message{
 			Command: "WHOIS",
 			Params:  params,
 		})
@@ -1288,7 +1288,7 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 		}
 		uc := ucChannel
 
-		uc.SendMessageLabeled(dc, &irc.Message{
+		uc.SendMessageLabeled(dc.id, &irc.Message{
 			Command: "INVITE",
 			Params:  []string{upstreamUser, upstreamChannel},
 		})
--- a/upstream.go
+++ b/upstream.go
@@ -59,6 +59,10 @@ type upstreamConn struct {
 
 	saslClient  sasl.Client
 	saslStarted bool
+
+	// set of LIST commands in progress, per downstream
+	// access is synchronized with user.pendingLISTsLock
+	pendingLISTDownstreamSet map[uint64]struct{}
 }
 
 func connectToUpstream(network *network) (*upstreamConn, error) {
@@ -79,19 +83,20 @@ func connectToUpstream(network *network) (*upstreamConn, error) {
 
 	outgoing := make(chan *irc.Message, 64)
 	uc := &upstreamConn{
-		network:               network,
-		logger:                logger,
-		net:                   netConn,
-		irc:                   irc.NewConn(netConn),
-		srv:                   network.user.srv,
-		user:                  network.user,
-		outgoing:              outgoing,
-		channels:              make(map[string]*upstreamChannel),
-		caps:                  make(map[string]string),
-		batches:               make(map[string]batch),
-		availableChannelTypes: stdChannelTypes,
-		availableChannelModes: stdChannelModes,
-		availableMemberships:  stdMemberships,
+		network:                  network,
+		logger:                   logger,
+		net:                      netConn,
+		irc:                      irc.NewConn(netConn),
+		srv:                      network.user.srv,
+		user:                     network.user,
+		outgoing:                 outgoing,
+		channels:                 make(map[string]*upstreamChannel),
+		caps:                     make(map[string]string),
+		batches:                  make(map[string]batch),
+		availableChannelTypes:    stdChannelTypes,
+		availableChannelModes:    stdChannelModes,
+		availableMemberships:     stdMemberships,
+		pendingLISTDownstreamSet: make(map[uint64]struct{}),
 	}
 
 	go func() {
@@ -136,6 +141,8 @@ func (uc *upstreamConn) Close() error {
 		return fmt.Errorf("upstream connection already closed")
 	}
 	close(uc.closed)
+
+	uc.endPendingLists(true)
 	return nil
 }
 
@@ -170,6 +177,81 @@ func (uc *upstreamConn) isChannel(entity string) bool {
 		return true
 	}
 	return false
+}
+
+func (uc *upstreamConn) getPendingList() *pendingLIST {
+	uc.user.pendingLISTsLock.Lock()
+	defer uc.user.pendingLISTsLock.Unlock()
+	for _, pl := range uc.user.pendingLISTs {
+		if _, ok := pl.pendingCommands[uc.network.ID]; !ok {
+			continue
+		}
+		return &pl
+	}
+	return nil
+}
+
+func (uc *upstreamConn) endPendingLists(all bool) (found bool) {
+	found = false
+	uc.user.pendingLISTsLock.Lock()
+	defer uc.user.pendingLISTsLock.Unlock()
+	for i := 0; i < len(uc.user.pendingLISTs); i++ {
+		pl := uc.user.pendingLISTs[i]
+		if _, ok := pl.pendingCommands[uc.network.ID]; !ok {
+			continue
+		}
+		delete(pl.pendingCommands, uc.network.ID)
+		if len(pl.pendingCommands) == 0 {
+			uc.user.pendingLISTs = append(uc.user.pendingLISTs[:i], uc.user.pendingLISTs[i+1:]...)
+			i--
+			uc.forEachDownstreamByID(pl.downstreamID, func(dc *downstreamConn) {
+				dc.SendMessage(&irc.Message{
+					Prefix:  dc.srv.prefix(),
+					Command: irc.RPL_LISTEND,
+					Params:  []string{dc.nick, "End of /LIST"},
+				})
+			})
+		}
+		found = true
+		if !all {
+			delete(uc.pendingLISTDownstreamSet, pl.downstreamID)
+			uc.user.forEachUpstream(func(uc *upstreamConn) {
+				uc.trySendList(pl.downstreamID)
+			})
+			return
+		}
+	}
+	return
+}
+
+func (uc *upstreamConn) trySendList(downstreamID uint64) {
+	// must be called with a lock in uc.user.pendingLISTsLock
+
+	if _, ok := uc.pendingLISTDownstreamSet[downstreamID]; ok {
+		// a LIST command is already pending
+		// we will try again when that command is completed
+		return
+	}
+
+	for _, pl := range uc.user.pendingLISTs {
+		if pl.downstreamID != downstreamID {
+			continue
+		}
+		// this is the first pending LIST command list of the downstream
+		listCommand, ok := pl.pendingCommands[uc.network.ID]
+		if !ok {
+			// there is no command for this upstream in these LIST commands
+			// do not send anything
+			continue
+		}
+		// there is a command for this upstream in these LIST commands
+		// send it now
+
+		uc.SendMessageLabeled(downstreamID, listCommand)
+
+		uc.pendingLISTDownstreamSet[downstreamID] = struct{}{}
+		return
+	}
 }
 
 func (uc *upstreamConn) parseMembershipPrefix(s string) (membership *membership, nick string) {
@@ -833,6 +915,29 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 			return fmt.Errorf("failed to parse topic time: %v", err)
 		}
 		ch.TopicTime = time.Unix(sec, 0)
+	case irc.RPL_LIST:
+		var channel, clients, topic string
+		if err := parseMessageParams(msg, nil, &channel, &clients, &topic); err != nil {
+			return err
+		}
+
+		pl := uc.getPendingList()
+		if pl == nil {
+			return fmt.Errorf("unexpected RPL_LIST: no matching pending LIST")
+		}
+
+		uc.forEachDownstreamByID(pl.downstreamID, func(dc *downstreamConn) {
+			dc.SendMessage(&irc.Message{
+				Prefix:  dc.srv.prefix(),
+				Command: irc.RPL_LIST,
+				Params:  []string{dc.nick, dc.marshalChannel(uc, channel), clients, topic},
+			})
+		})
+	case irc.RPL_LISTEND:
+		ok := uc.endPendingLists(false)
+		if !ok {
+			return fmt.Errorf("unexpected RPL_LISTEND: no matching pending LIST")
+		}
 	case irc.RPL_NAMREPLY:
 		var name, statusStr, members string
 		if err := parseMessageParams(msg, nil, &statusStr, &name, &members); err != nil {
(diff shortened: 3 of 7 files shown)


upstream.go
```diff
@@ -860,3 +1128,14 @@ func (uc *upstreamConn) readMessages(ch chan<- upstreamIncomingMessage) error {
 func (uc *upstreamConn) SendMessage(msg *irc.Message) {
 	uc.outgoing <- msg
 }
+
+func (uc *upstreamConn) SendMessageLabeled(dc *downstreamConn, msg *irc.Message) {
+	if uc.labelsSupported {
+		if msg.Tags == nil {
+			msg.Tags = make(map[string]irc.TagValue)
+		}
+		msg.Tags["label"] = irc.TagValue(fmt.Sprintf("sd-%d-%d", dc.id, uc.nextLabelId))
```
delthas: Should this be obfuscated somehow? (I think not). Also, are uint64 counters OK for downstream ids and label ids? Either that or randomly generated UUIDs could work.

upstream.go
```diff
@@ -697,11 +958,14 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 		}
 
 		uc.forEachDownstream(func(dc *downstreamConn) {
+			if downstreamId != 0 && downstreamId != dc.id {
+				return
+			}
```
delthas: This piece of code used to select a specific downstream id is used multiple times throughout the code, should we create a function for that? Maybe a new `forEachDownstream` that also takes an optional downstream id.

irc.go
```diff
@@ -54,6 +54,89 @@ func (ms *modeSet) Apply(s string) error {
...
+	'l': modeTypeC,
+	'i': modeTypeD,
+	'm': modeTypeD,
+	'n': modeTypeD,
+	's': modeTypeD,
+	't': modeTypeD,
+}
+
+func ApplyChannelModes(modeTypes map[byte]channelModeType, modes map[byte]string, modeStr string, arguments ...string) error {
+	nextArgument := 0
```
emersion: We should use a consistent naming: either argument or param. Here we have `arguments`, `param`, `nextArgument`.

irc.go
```diff
@@ -54,6 +54,89 @@ func (ms *modeSet) Apply(s string) error {
...
+		}
+		if plusMinus != '+' && plusMinus != '-' {
+			return fmt.Errorf("malformed modestring %q: missing plus/minus", modeStr)
+		}
+
+		mt, ok := modeTypes[mode]
+		if !ok {
+			continue
+		}
+		if mt == modeTypeB || (mt == modeTypeC && plusMinus == '+') {
```
emersion: This can probably be simplified to something among those lines:

```go
if plusMinus == '+' {
	var param string
	switch mt {
	case modeTypeB, modeTypeC:
		// some sentitive arguments […]
		if nextArgument < len(arguments) {
			…
		}
	case modeTypeC, modeTypeD:
		// Mode has no param
	}
	modes[mode] = param
} else {
	delete(modes, mode)
}
```

irc.go
```diff
@@ -54,6 +54,89 @@ func (ms *modeSet) Apply(s string) error {
 	return nil
 }
 
+type channelModeType byte
+
+const (
+	modeTypeA channelModeType = iota
```
emersion: Would be nice to have doc comments about what these are

irc.go
```diff
@@ -54,6 +54,89 @@ func (ms *modeSet) Apply(s string) error {
...
+	'k': modeTypeB,
+	'l': modeTypeC,
+	'i': modeTypeD,
+	'm': modeTypeD,
+	'n': modeTypeD,
+	's': modeTypeD,
+	't': modeTypeD,
+}
+
+func ApplyChannelModes(modeTypes map[byte]channelModeType, modes map[byte]string, modeStr string, arguments ...string) error {
```
emersion: Just like we have `userModes`, maybe it makes sense to have `type channelModes map[byte]string` with methods like `Apply` and `String`?

irc.go
delthas: Right. I've picked argument, for "mode argument" (as written in the specs)

irc.go
delthas: Not sure I can simply further. See `nextArgument++`.

irc.go
delthas: Added.

irc.go
delthas: OK. I'm not sure about "`String()`" though b/c we don't return a string but also a string array.

emersion: What about `Format` or `Marshal`?

delthas: `Format` is good, i'll do that

upstream.go
```diff
@@ -508,6 +548,88 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+		if err != nil {
+			return err
+		}
+
+		firstMode := ch.modes == nil
+		ch.modes = make(map[byte]string)
+		if err := ch.modes.Apply(uc.availableChannelModes, modeStr, msg.Params[3:]...); err != nil {
+			return err
+		}
+		if firstMode {
```
emersion: We should probably send `RPL_CHANNELMODEIS` to downstream connections unconditionally. Otherwise clients may become out-of-sync.

upstream.go
```diff
@@ -508,6 +548,88 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+		}
+		modeStr := ""
+		if len(msg.Params) > 1 {
+			modeStr = msg.Params[1]
+		}
+
+		uc.modes = ""
+		if err := uc.modes.Apply(modeStr); err != nil {
+			return err
+		}
```
emersion: Can we add a TODO: send RPL_UMODEIS to downstream connections when applicable?

downstream.go
```diff
@@ -913,6 +879,49 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 				})
 			}
```
emersion: A `break` is missing here, to prevent the `!uc.isChannel(upstreamName)` check to send an error.

upstream.go
```diff
@@ -346,11 +325,67 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+				negate = true
+				token = token[1:]
+			} else {
+				if i := strings.IndexByte(token, '='); i >= 0 {
+					parameter = token[:i]
+					value = token[i+1:]
+				}
+			}
+			if !negate {
+				switch parameter {
```
emersion: Calling `strings.ToUpper` would allow more liberal parsing.

downstream.go
```diff
@@ -891,45 +891,75 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
...
-			})
 		} else {
-			ch, ok := uc.channels[upstreamName]
-			if !ok {
-				return ircError{&irc.Message{
-					Command: irc.ERR_NOSUCHCHANNEL,
-					Params:  []string{dc.nick, name, "No such channel"},
-				}}
-			}
+			if modeStr != "" {
```
emersion: This change is unrelated to this commit. I dropped it while applying the patch to master.

downstream.go
```diff
@@ -891,45 +891,75 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
...
+		}
+
+		channels := strings.Split(msg.Params[0], ",")
+		for _, channel := range channels {
+			uc, upstreamChannel, err := dc.unmarshalEntity(channel)
+			if err != nil {
+				return err
+			}
+
+			ch, err := uc.getChannel(upstreamChannel)
```
emersion: I prefer to use `ch, ok := uc.channels[upstreamChannel]` when we want to handle both cases (the case where the channel exists, and the case where it doesn't). `getChannel` is just a helper to unify "channel not found" errors.

emersion: (TBH we could replace `getChannel` with an `errNoSuchChannel` variable)

upstream.go
```diff
@@ -757,7 +757,7 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 			dc.SendMessage(&irc.Message{
 				Prefix:  dc.srv.prefix(),
 				Command: irc.RPL_ENDOFWHO,
-				Params:  []string{dc.nick, name, "End of WHO list"},
+				Params:  []string{dc.nick, name, "End of /WHO list"},
```
emersion: Unrelated change, moving it to an earlier commit while cherry-picking.

upstream.go
```diff
@@ -984,6 +1002,8 @@ func (uc *upstreamConn) handleCapAck(name string, ok bool) error {
 			Command: "AUTHENTICATE",
 			Params:  []string{auth.Mechanism},
 		})
+	case "message-tags":
+		uc.tagsSupported = true
```
emersion: `= ok`

upstream.go
```diff
@@ -888,6 +904,8 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 				Params:  []string{dc.marshalNick(uc, nick), dc.marshalChannel(uc, channel)},
 			})
 		})
+	case "TAGMSG":
+		// Ignore
```
emersion: The spec says this should be handled just like a `PRIVMSG` (ie. relayed). We should add a `// TODO: relay to downstream accepting message-tags`.

emersion:
Cherry-picked all commits up to (and including) "Add downstream self WHO and WHOIS support".

To rebase:

```sh
git fetch upstream
git rebase -i upstream/master
# pick → drop all commits cherry-picked in master
```

upstream.go
emersion: Yeah, I think all of this is fine.

upstream.go
```diff
@@ -1044,6 +1074,8 @@ func (uc *upstreamConn) handleCapAck(name string, ok bool) error {
 		})
 	case "message-tags":
 		uc.tagsSupported = true
+	case "labeled-response":
+		uc.labelsSupported = true
```
emersion: `= ok`

upstream.go
emersion: +1

upstream.go
hhirtz: Here, you can apply this suggestion ;)

```suggestion
		// TODO: relay to downstream accepting message-tags
```

emersion: Please don't use GitHub suggestions. Those add fixup commits instead of amending the original commit.

hhirtz: Oh ok, didn't know.  I see how this is not wanted here.

emersion:
Pushed all commits except the last one. Thanks!

upstream.go
```diff
@@ -206,9 +206,28 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 	case "NOTICE":
 		uc.logger.Print(msg)
 
-		uc.forEachDownstream(func(dc *downstreamConn) {
-			dc.SendMessage(msg)
-		})
+		if msg.Prefix.User == "" && msg.Prefix.Host == "" {
+			uc.SendMessage(msg)
```
emersion: This bounces the message back to the upstream server. We probably want to send it to downstream instead, or am I missing something?

upstream.go
```diff
@@ -206,9 +206,28 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+			var nick, text string
+			if err := parseMessageParams(msg, &nick, &text); err != nil {
+				return err
+			}
+
+			target := nick
+			if nick == uc.nick {
+				target = msg.Prefix.Name
+			}
+			uc.Logfn(target, "<%s> %s", msg.Prefix.Name, text)
```
emersion: This belongs to a later commit.

emersion: Ah, it seems like this has been removed in 3d8819ad8054f25aa83a83a23208f5a70c0b619f.

downstream.go
```diff
@@ -1236,6 +1236,23 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
...
+		}
+
+		for _, name := range strings.Split(targetsStr, ",") {
+			uc, upstreamName, err := dc.unmarshalEntity(name)
+			if err != nil {
+				return err
+			}
+
+			uc.SendMessage(&irc.Message{
+				Command: "PRIVMSG",
```
emersion: Shouldn't this be `NOTICE`?

user.go
```diff
@@ -87,6 +87,15 @@ type user struct {
 	networks        []*network
 	nextUpstreamID  uint64 // use with atomic
 	downstreamConns []*downstreamConn
+
+	pendingListsLock sync.Mutex
```
emersion: Note to self: all messages (upstream and downstream) are processed in a per-user goroutine (I should write down the goroutine architecture). This means that the only reason why we need this lock is to clear the pending lists when an upstream connection is closed.

upstream.go
```diff
@@ -1089,6 +1211,8 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 		// Ignore
 	case irc.RPL_MOTDSTART, irc.RPL_MOTD, irc.RPL_ENDOFMOTD:
 		// Ignore
+	case irc.RPL_LISTSTART:
```
emersion: Do we need to send an initial `RPL_LISTSTART` to clients?

delthas: Probably not, it's optional and marked as deprecated in RFC 2812

upstream.go
delthas: Right, thats a mistake

downstream.go
delthas: Right!

upstream.go
```diff
@@ -204,9 +204,30 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 	case "NOTICE":
 		uc.logger.Print(msg)
 
-		uc.forEachDownstream(func(dc *downstreamConn) {
-			dc.SendMessage(msg)
-		})
+		if msg.Prefix.User == "" && msg.Prefix.Host == "" { // server message
```
emersion: Can the IRC server send a NOTICE without any prefix? This is what Freenode sends for a server NOTICE:

```
:verne.freenode.net NOTICE * :*** Looking up your hostname...
```

upstream.go
```diff
@@ -204,9 +204,30 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+				dc.SendMessage(msg)
+			})
+		} else { // regular user NOTICE
+			var nick, text string
+			if err := parseMessageParams(msg, &nick, &text); err != nil {
+				return err
+			}
+
+			target := nick
+			if nick == uc.nick {
```
emersion: Why do we have this special case? If the target is ourselves, why should we change the target to the source of the NOTICE message?

emersion: Oh, this is only used for logs. Can we move this whole code block in the commit adding support for logs? (`git rebase -i`, edit both commits, cut the block from this commit, `git rebase --continue`, paste the block to the next one)

config/config.go
```diff
@@ -19,18 +20,25 @@ type Server struct {
...
 	SQLSource string
+	LogPath   string
 }
 
 func Defaults() *Server {
 	hostname, err := os.Hostname()
 	if err != nil {
 		hostname = "localhost"
 	}
+	logPath, err := os.UserHomeDir()
```
emersion: Not sure about this default. soju is designed to be multi-user, so logging in ~ sounds weird. I'd prefer having logs disabled by default and enabled in the config file.

upstream.go
```diff
@@ -1344,3 +1395,47 @@ func (uc *upstreamConn) SendMessageLabeled(downstreamID uint64, msg *irc.Message
 	}
 	uc.SendMessage(msg)
 }
+
+// TODO: handle moving logs when a network name changes, when support for this is added
+func (uc *upstreamConn) Logfn(entity string, format string, a ...interface{}) {
```
emersion: Maybe we could rename this to `AppendLog` or `WriteLog`? The `fn` suffix doesn't add a lot of value.

upstream.go
emersion: Ugh, GitHub shows these lines as added in both the NOTICE commit and the logging commit. Not sure why.

upstream.go
```diff
@@ -603,6 +622,8 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 				delete(ch.Members, msg.Prefix.Name)
 				ch.Members[newNick] = membership
 			}
+
+			uc.Logfn(ch.Name, "*** %s is now known as %s", msg.Prefix.Name, newNick)
```
emersion: We should only log this if the user is part of the channel members.

emersion:
Merged "Add downstream NOTICE support".

Apart from the comments, the meat of the "Add support for bouncer logs" patch looks good to me!

upstream.go
delthas: in that case, `msg.Prefix.Name = verne.freenode.net`, `msg.Prefix.User = ""`, `msg.Prefix.Host = ""`. that's why i used `msg.Prefix.User == "" && msg.Prefix.Host == ""`

config/config.go
delthas: Right, makes sense

upstream.go
delthas: OK,  `AppendLog` is fine

upstream.go
delthas: Right

upstream.go
delthas: Should be fixed now.

upstream.go
emersion: Ah, got it!

upstream.go
```diff
@@ -1438,3 +1437,11 @@ func (uc *upstreamConn) AppendLog(entity string, format string, a ...interface{}
 		uc.logger.Printf("failed to log message to %q: %v", log.name, err)
 	}
 }
+
+func splitSpace(s string) []string {
```
emersion: Maybe we can just use `strings.Fields`?

upstream.go
```diff
@@ -725,12 +756,19 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+		}
+
 		if msg.Prefix.Name == uc.nick {
 			uc.logger.Printf("quit")
 		}
 
 		for _, ch := range uc.channels {
 			delete(ch.Members, msg.Prefix.Name)
+
+			uc.AppendLog(ch.Name, "*** Quits: %s (%s@%s) (%s)", msg.Prefix.Name, msg.Prefix.User, msg.Prefix.Host, reason)
```
emersion: We need to check whether the user belongs to the channel, otherwise we'll append this to all channel logs.

downstream.go
```diff
@@ -1077,6 +1077,46 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
...
+		var upstreamChannels map[int64][]string
+		if len(msg.Params) > 0 {
+			upstreamChannels = make(map[int64][]string)
+			channels := strings.Split(msg.Params[0], ",")
+			for _, channel := range channels {
+				uc, upstreamChannel, err := dc.unmarshalEntity(channel)
+				if err != nil {
+					return err
+				}
+				if c, ok := upstreamChannels[uc.network.ID]; ok {
```
emersion: Can be simplified to:

```go
upstreamChannels[uc.network.ID] = append(upstreamChannels[uc.network.ID], upstreamChannel)
```

upstream.go
```diff
@@ -155,6 +162,84 @@ func (uc *upstreamConn) isChannel(entity string) bool {
...
+					Prefix:  dc.srv.prefix(),
+					Command: irc.RPL_LISTEND,
+					Params:  []string{dc.nick, "End of /LIST"},
+				})
+			})
+		}
+		found = true
+		if !all {
+			delete(uc.pendingListDownstreamSet, pl.downstreamID)
+			uc.user.forEachUpstream(func(uc *upstreamConn) {
```
emersion: Since this function can only "unblock" a LIST command for the current upstream connection `uc`, do we really need to iterate over all upstream connections here?

delthas: Suppose we're connected to downstreams *a*, *b*; and the client does LIST twice.

Pending now schematically looks like this: `[[a, b], [a, b]]`. Now imagine *a* answers much faster than *b*. After *a* answers, we get: `[[b], [a, b]]`. We can't yet send the other LIST to *a* because it would be sent before the first RPL_LISTEND. When *b* finally replies, we get: `[[], [a, b]]`, ie `[[a, b]]`. We now need to resend both  `a` and `b` for the next LIST.

Hence the need to iterate over all upstreams.

user.go
```diff
@@ -96,6 +96,15 @@ type user struct {
...
+	pendingLists     []pendingList
+}
+
+type pendingList struct {
+	downstreamID uint64
+	// list of pending per-upstream LIST requests and their corresponding LIST message parameters
+	pendingRequests map[int64][]string
```
emersion: Can we replace `[]string` with the whole `*irc.Message`? It would make it clearer what the items are.

Might as well rename `pendingRequests` to `pendingCommands`?

user.go
```diff
@@ -96,6 +96,15 @@ type user struct {
...
 	networks        []*network
 	downstreamConns []*downstreamConn
+
+	pendingListsLock sync.Mutex
+	pendingLists     []pendingList
+}
+
+type pendingList struct {
+	downstreamID uint64
+	// list of pending per-upstream LIST requests and their corresponding LIST message parameters
```
emersion:
```go
// list of per-upstream LIST commands not yet sent
```

user.go
```diff
@@ -96,6 +96,15 @@ type user struct {
 
 	networks        []*network
 	downstreamConns []*downstreamConn
+
+	pendingListsLock sync.Mutex
+	pendingLists     []pendingList
```
emersion:
```go
// LIST commands in progress
```

user.go
delthas: Right, I found it confusing as well, at first I put a typedef for that but it was kinda overkill. `*irc.Message` is a good idea.

user.go
emersion: Maybe we could even rename this to `pendingLISTs` to make it clear it's about LIST, and not just a list?

user.go
delthas: `not yet complete`*? we only clear it when we receive the last RPL_LISTEND (or error, or close)

downstream.go
```diff
@@ -1077,6 +1077,46 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
...
+					upstreamChannels[uc.network.ID] = []string{upstreamChannel}
+				}
+			}
+		}
+
+		dc.user.pendingLists = append(dc.user.pendingLists, pl)
+		dc.forEachUpstream(func(uc *upstreamConn) {
+			var params []string
+			if upstreamChannels == nil {
+				params = []string{}
```
emersion: Nit: no need to initialize `params`, `params = nil` would do (or just do nothing).

upstream.go
```diff
@@ -59,6 +59,10 @@ type upstreamConn struct {
 
 	saslClient  sasl.Client
 	saslStarted bool
+
+	// set of pending LIST requests, per downstream
```
emersion:
```go
// set of LIST commands in progress, per downstream
```

to make it clear what "pending" means (sent but not finished, as opposed to not yet sent)

user.go
emersion: Eh, right! +1

PR #9, opened 2020-04-04
Multiple features

delthas:
_Several commits in one PR for easier rebasing. See individual commits for details._

_Not that Github may show some commits in the wrong order (ordered by date rather than by parent)._

### DIFF
--- a/service.go
+++ b/service.go
@@ -12,6 +12,12 @@ import (
 
 const serviceNick = "BouncerServ"
 
+var servicePrefix = &irc.Prefix{
+	Name: serviceNick,
+	User: serviceNick,
+	Host: serviceNick,
+}
+
 type serviceCommandSet map[string]*serviceCommand
 
 type serviceCommand struct {
@@ -23,7 +29,7 @@ type serviceCommand struct {
 
 func sendServiceNOTICE(dc *downstreamConn, text string) {
 	dc.SendMessage(&irc.Message{
-		Prefix:  &irc.Prefix{Name: serviceNick},
+		Prefix:  servicePrefix,
 		Command: "NOTICE",
 		Params:  []string{dc.nick, text},
 	})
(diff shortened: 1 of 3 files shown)


service.go
```diff
@@ -199,7 +207,16 @@ func handleServiceNetworkStatus(dc *downstreamConn, params []string) error {
...
 			details = fmt.Sprintf("%v channels", len(uc.channels))
 		} else {
-			statuses = append(statuses, "disconnected")
+			var lastError error
+			net.lock.Lock()
+			lastError = net.lastError
+			net.lock.Unlock()
+
+			if lastError != nil {
+				statuses = append(statuses, fmt.Sprintf("disconnected (%v)", lastError))
```
emersion, line 222: Maybe we can use `details` to not clutter the statuses wit the error message?

service.go
```diff
@@ -199,7 +207,16 @@ func handleServiceNetworkStatus(dc *downstreamConn, params []string) error {
 			statuses = append(statuses, "connected as "+uc.nick)
 			details = fmt.Sprintf("%v channels", len(uc.channels))
 		} else {
-			statuses = append(statuses, "disconnected")
+			var lastError error
+			net.lock.Lock()
```
emersion, line 217: The service commands are handled from the user goroutine, so no lock should be required.

user.go
```diff
@@ -43,8 +54,9 @@ type network struct {
 
 	history map[string]uint64
 
-	lock sync.Mutex
-	conn *upstreamConn
+	lock      sync.Mutex
+	conn      *upstreamConn
+	lastError error
```
emersion, line 59: (If this doesn't require a lock, let's move it above `lock`, alongside `history`)

user.go
```diff
@@ -57,6 +69,15 @@ func newNetwork(user *user, record *Network) *network {
 	}
 }
 
+func (net *network) forEachDownstream(f func(*downstreamConn)) {
```
emersion, line 72: Maybe we can change `upstreamConn.forEachDownstream` to use this?

delthas:
Updated.

PR #10, opened 2020-04-15
Add support for custom network on-connect messages

delthas:
Some servers use custom IRC bots with custom commands for registering to
specific services after connection.

This adds support for setting a custom raw IRC message, that will be
sent after registering to a network.

### DIFF
--- a/service.go
+++ b/service.go
@@ -104,7 +104,7 @@ func init() {
 		"network": {
 			children: serviceCommandSet{
 				"create": {
-					usage:  "-addr <addr> [-name name] [-username username] [-pass pass] [-realname realname] [-nick nick]",
+					usage:  "-addr <addr> [-name name] [-username username] [-pass pass] [-realname realname] [-nick nick] [[-connect-command command] ...]",
 					desc:   "add a new network",
 					handle: handleServiceCreateNetwork,
 				},
@@ -174,6 +174,17 @@ func newFlagSet() *flag.FlagSet {
 	return fs
 }
 
+type stringSliceVar []string
+
+func (v *stringSliceVar) String() string {
+	return fmt.Sprint([]string(*v))
+}
+
+func (v *stringSliceVar) Set(s string) error {
+	*v = append(*v, s)
+	return nil
+}
+
 func handleServiceCreateNetwork(dc *downstreamConn, params []string) error {
 	fs := newFlagSet()
 	addr := fs.String("addr", "", "")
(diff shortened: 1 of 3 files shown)


delthas:
@emersion : thoughts? This feature is important to me but I'm not sure how to implement it (a flag in `network create`? a custom service command? should we support a list of messages or just one? wording (is `-connect-message` ok?))

Also of note that since we don't have `network modify` yet, setting the connect message would require deleting and creating the network again, but after `network modify` is added this should be a bit less tedious.

delthas:
By the way, excellent job on the db migration design, I can't belive adding a column is as easy as adding an `alter table` statement in the migrations array. 👍

emersion:
I'm not against the idea, sadly some networks still don't support SASL. Note that SASL is superior because it provides better error handling and authenticates before registration completes (NickServ causes issues when the user wants to join restricted channels for instance, like #freebsd on Freenode).

Because this is more-or-less legacy support, I'd prefer not to do anything too complicated to support this feature.

Are there other potential use-cases than authentication?

Maybe a better name would be `post_register_command`? (Commands sent after registration is complete) But maybe that would just be more confusing from a user PoV, not sure.

>This feature is important to me but I'm not sure how to implement it (a flag in network create? a custom service command?

A network flag sounds best.

>should we support a list of messages or just one? 

Maybe store messages separated by `\r\n` in the DB, and allow the user to configure them via multiple flags? (e.g. `-connect-command "PRIVMSG nick1 hey there" -connect-command "PRIVMSG nick2 hey there"`) But then how to unset the list? Or maybe require a literal `\n` (`-connect-command "PRIVMSG nick1 hey there\nRIVMSG nick2 hey there"`)

delthas:
> Are there other potential use-cases than authentication?
Yes-ish: some servers or chanels use a custom IRC bot, distinct from NickServ, that you send a privmsg to in order to join a channel.

> Maybe store messages separated by \r\n in the DB, and allow the user to configure them via multiple flags?
👍 

> But then how to unset the list?
`-connect-command ""` (same as other fields)

emersion:
This plan looks good to me.

delthas:
Updated. Note that I've put the string split/join in the db methods to "abstract away" its storage as a `\r\n` string. The other solution is to split it each time when sending it after registering. I think it doesn't really matter performance-wise.

service.go
```diff
@@ -174,6 +174,17 @@ func newFlagSet() *flag.FlagSet {
 	return fs
 }
 
+type StringSliceVar []string
```
emersion, line 177: Let's keep this type unexported

PR #11, opened 2020-04-18
Add support for skipping TLS certificate verification

delthas:
Some servers use expired TLS certificates. Connecting to these networks
is only possible by skipping TLS certificate verification. This feature
is also useful for connecting to servers with different hostnames
(although a specific option for setting the right hostname could be
added in the future).

This adds support for a flag in the service command `network create`
that lets users disable TLS certificate verification for a particular
network.

### DIFF
--- a/db.go
+++ b/db.go
@@ -33,6 +33,7 @@ type Network struct {
 	Pass            string
 	ConnectCommands []string
 	SASL            SASL
+	InsecureTLS     bool
 }
 
 func (net *Network) GetName() string {
@@ -69,6 +70,7 @@ CREATE TABLE Network (
 	sasl_mechanism VARCHAR(255),
 	sasl_plain_username VARCHAR(255),
 	sasl_plain_password VARCHAR(255),
+	insecure_tls INTEGER,
 	FOREIGN KEY(user) REFERENCES User(username),
 	UNIQUE(user, addr, nick)
 );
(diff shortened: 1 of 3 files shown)


delthas:
context: I do need a way to access servers with invalid certificated, it is an actual use case for me. Some of them have the wrong hostname, some of them have expired. This PR adds a simple flag and few code for this feature.

We could also make a list of allowed fingerprints like ZNC does but this would require a bit more code and would be more tedious from a user POV. As for my threat model, an MITM on the IRC servers that have invalid certs is very very unlikely. I'd rather not bother updating the fingerprints every 3 months.

emersion:
I'd prefer an option to accept a specific TLS certificate fingerprint.

emersion:
Disabling TLS makes the connection as good as a plain-text one. I'd rather add support for insecure plain-text connections instead.

delthas:
I kinda disagree; it's hugely more likely that packets are passively read by a third party (in which case an encrypted connection is much better than a plain-text one), rather than the connection rerouted to an attacker with a different certificate. Accepting invalid certificates is much better in the 99.9% cases  where some firewall or attacker spies on the connection without rerouting it and presenting an invalid certificate.

Especially more so in a context where soju runs on a server in a datacenter. Is it really likely at all that the connection be rerouted to an attacker? It's actually much more likely that some attacker takes over the IRC server, its cert, and gets access to all messages without changing the cert.

Ultimately this is an opiniated decision so I'll accept your choice but I do believe accepting insecure TLS certificates rather than falling back to plain text is much more secure in the large majority of cases.

emersion:
This threat model is wrong. If someone can MITM to read the plain-text traffic, they can also trivially MITM to mutate it.

I don't want to offer users a way to setup an "insecure TLS" connection. Either it's insecure plain-text, either it's secure TLS. Having a it's-TLS-but-it's-insecure is just confusing.

emersion:
>Some servers use expired TLS certificates

BTW, the right fix is to pester the IRC server maintainer to keep an up-to-date cert, like everybody else does.

delthas:
OK. See #12 for plaintext upstreams.

PR #12, opened 2020-04-19
Add support for plaintext upstreams (-no-tls)

delthas:
Some servers do not support TLS, or have invalid, expired, or
self-signed TLS certificates. While the right fix would be to contact
each server owner to add support for valid TLS, supporting plaintext
upstream connections is sometimes necessary.

This adds support for a flag in the service command `network create`
that lets users disable TLS for a particular network.

### DIFF
--- a/db.go
+++ b/db.go
@@ -33,6 +33,7 @@ type Network struct {
 	Pass            string
 	ConnectCommands []string
 	SASL            SASL
+	NoTLS           bool
 }
 
 func (net *Network) GetName() string {
@@ -69,6 +70,7 @@ CREATE TABLE Network (
 	sasl_mechanism VARCHAR(255),
 	sasl_plain_username VARCHAR(255),
 	sasl_plain_password VARCHAR(255),
+	no_tls INTEGER,
 	FOREIGN KEY(user) REFERENCES User(username),
 	UNIQUE(user, addr, nick)
 );
@@ -86,6 +88,7 @@ CREATE TABLE Channel (
 var migrations = []string{
 	"", // migration #0 is reserved for schema initialization
 	"ALTER TABLE Network ADD COLUMN connect_commands VARCHAR(1023)",
+	"ALTER TABLE Network ADD COLUMN no_tls INTEGER DEFAULT 0",
 }
 
 type DB struct {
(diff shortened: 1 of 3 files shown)


emersion:
Hmm. I wonder if it would make more sense to accept a full URL in `-addr` instead of adding a flag. That would allow us to support things like `irc.example.org` (current syntax), `ircs://irc.example.org` (same), `irc+insecure://irc.example.org` (plain-text), `unix:///run/bitlbee.socket` (Unix socket). This is something done by aerc and koushin.

delthas:
Hmm... I don't see any use case that requires Unix socket. For example connecting to bitlbee could be done by TCP as well. I guess it could also be a bit more confusing for users to search for the right URL protocol for IRC without TLS, rather than simply passing a `no-tls` flag.

But both solutions are fine to me, URLs also kinda make sense. Let's settle on the solution before I make the PR though so I don't have to open yet another PR if we choose another solution 😛

emersion:
>I don't see any use case that requires Unix socket. For example connecting to bitlbee could be done by TCP as well.

Unix sockets allow:

- To make sure the socket isn't expose to the outside world, even if the bind address or firewall is misconfigured
- To use Unix file permissions to restrict access to the socket
- Unix sockets are less complex, because they don't require a whole network stack

>I guess it could also be a bit more confusing for users to search for the right URL protocol for IRC without TLS, rather than simply passing a no-tls flag.

That's true. However I'm not too worried about making it non-immediately-obvious to add a non-TLS server.

So, on the whole, I think I still prefer to use URLs.

delthas:
OK, makes sense.

Should `irc://` be plaintext or TLS? [draft-mirashi-url-irc-01](https://tools.ietf.org/html/draft-mirashi-url-irc-01) and [draft-butcher-irc-url-04](https://tools.ietf.org/html/draft-butcher-irc-url-04) say `irc://` should be plaintext. Other implementations I'm aware of also se plaintext for `irc`.

I'd rather have `irc://` for plaintext and `ircs://` for TLS, defaulting to TLS when there's no URL scheme.

Do we really need `irc+insecure://`?

emersion:
>Do we really need irc+insecure://?

Yes.

delthas:
See #15 .

PR #14, opened 2020-04-21
Unmarshal nicks in texts of PRIVMSG and NOTICE from downstreams

delthas:
When writing a PRIVMSG or NOTICE on a channel, it is very common to use
autocompletion to mention other users on that channel. When using soju
in multi-network mode, all users will have their nicked suffixed by
`/network`. This suffix should be removed before sending it upstream.

This adds support for removing all `/network` suffixes in messages sent
to a channel of that network.

### DIFF
--- a/downstream.go
+++ b/downstream.go
@@ -199,6 +199,14 @@ func (dc *downstreamConn) unmarshalEntity(name string) (*upstreamConn, string, e
 	return conn, name, nil
 }
 
+func (dc *downstreamConn) unmarshalText(uc *upstreamConn, text string) string {
+	if dc.upstream() != nil {
+		return text
+	}
+	// TODO: smarter parsing that ignores URLs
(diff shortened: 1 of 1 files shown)


delthas:
Very useful to me. Adding some smarter parsing later would be nice, to avoid URLs being modified if they happen to contain `/network` in their path.

downstream.go
```diff
@@ -546,6 +546,14 @@ func unmarshalUsername(rawUsername string) (username, client, network string) {
 	return username, client, network
 }
 
+func (dc *downstreamConn) unmarshalNicks(uc *upstreamConn, text string) string {
```
emersion, line 549: Can we rename this to e.g. `unmarshalText`? `unmarshalNicks` doesn't make it clear this accepts an arbitrary string.

emersion, line 549: Can we move this right under `unmarshalEntity`, since it's a related operation?

delthas:
Updated.

PR #15, opened 2020-04-27
Add support for IRC schemes: ircs, irc+insecure

delthas:

### DIFF
--- a/service.go
+++ b/service.go
@@ -203,6 +203,15 @@ func handleServiceCreateNetwork(dc *downstreamConn, params []string) error {
 		return fmt.Errorf("flag -addr is required")
 	}
 
+	if addrParts := strings.SplitN(*addr, "://", 2); len(addrParts) == 2 {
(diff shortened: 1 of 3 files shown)


service.go
```diff
@@ -203,6 +203,15 @@ func handleServiceCreateNetwork(dc *downstreamConn, params []string) error {
 		return fmt.Errorf("flag -addr is required")
 	}
 
+	if addrParts := strings.SplitN(*addr, "://", 2); len(addrParts) == 2 {
+		scheme := addrParts[0]
+		switch scheme {
+		case "ircs", "irc+insecure":
+		default:
+			return fmt.Errorf("unknown scheme %q: supported schemes: ircs, irc+insecure", scheme)
```
emersion, line 211: Nit: can we put the supported schemes in parens instead?

```go
"unknown scheme %q (supported schemes: ircs, irc+insecure)"
```

delthas:
Updated.

PR #16, opened 2020-04-30
Multiple features

delthas:

### DIFF
--- a/upstream.go
+++ b/upstream.go
@@ -1177,7 +1177,7 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 		})
 	case irc.RPL_INVITING:
 		var nick, channel string
-		if err := parseMessageParams(msg, &nick, &channel); err != nil {
+		if err := parseMessageParams(msg, nil, &nick, &channel); err != nil {
 			return err
 		}
 


downstream.go
```diff
@@ -57,6 +57,7 @@ var permanentDownstreamCaps = map[string]string{
 	"cap-notify":   "",
 	"echo-message": "",
 	"message-tags": "",
+	"multi-prefix": "",
```
emersion, line 60: Unfortunately we can't advertise multi-prefix unconditionally. We can only advertise it if all upstream connections support it.

See how away-notify is handled.

irc.go
```diff
@@ -176,11 +176,58 @@ var stdMemberships = []membership{
 	{'v', '+'}, // voice
 }
 
-func (m *membership) String() string {
-	if m == nil {
-		return ""
+// memberships always sorted by descending membership rank
+type memberships []membership
+
+func (m memberships) Add(availableMemberships []membership, newMembership membership) memberships {
```
emersion, line 215: Can we make this method have a pointer receiver (`func (m *memberships) Add(…)`)? It's a little too easy to ignore by mistake the return value of this function.

irc.go
```diff
@@ -176,11 +176,58 @@ var stdMemberships = []membership{
...
+		}
+	}
+	newMemberships := make([]membership, len(m)+1)
+	copy(newMemberships[:i], m[:i])
+	newMemberships[i] = newMembership
+	copy(newMemberships[i+1:], m[i:])
+	return newMemberships
+}
+
+func (m memberships) Remove(oldMembership membership) memberships {
```
emersion, line 240: Same

irc.go
```diff
@@ -176,11 +176,58 @@ var stdMemberships = []membership{
...
+			newMemberships := make([]membership, len(m)-1)
+			copy(newMemberships[:i], m[:i])
+			copy(newMemberships[i:], m[i+1:])
+			return newMemberships
+		}
+	}
+	return m
+}
+
+func (m memberships) String(dc *downstreamConn) string {
```
emersion, line 252: Now that this doesn't implement `fmt.Stringer` anymore, maybe we should rename this to `MarshalPrefixes`

irc.go
```diff
@@ -176,11 +176,58 @@ var stdMemberships = []membership{
...
+	return newMemberships
+}
+
+func (m memberships) Remove(oldMembership membership) memberships {
+	for i, currentMembership := range m {
+		if currentMembership == oldMembership {
+			newMemberships := make([]membership, len(m)-1)
+			copy(newMemberships[:i], m[:i])
+			copy(newMemberships[i:], m[i+1:])
+			return newMemberships
```
emersion, line 246: Can be simplified to `append(m[:i], m[i+1:]...)`

irc.go
delthas: what about `Format()`, like we do for channel modes?

delthas:
Updated.

emersion:
Hmm. I'm a little confused about `memberships.Remove`. I also understand why you didn't want to have a pointer receiver :P

Pushed a fixup commit with some suggestions, let me know what you think.

downstream.go
```diff
@@ -935,6 +944,15 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 		dc.forEachUpstream(func(uc *upstreamConn) {
 			uc.SendMessage(msg)
 		})
+
+		if dc.nick != nick {
```
emersion, line 948: Hmm. So when downstream sends a NICK command, this code sends back an ack, but just for this downstream connection. I think we should just wait for the server ack instead?

delthas: I'd also like the downstream nick to change if i'm connected to no upstreams or if i'm connected to multiple upstreams in multi-server mode. Right now the code in upstream.go only updates nicks when in single-server mode.

If you'd like we could change the downstream nick only if `dc.upstream() != nil` so that we wait for the server ack in that specific case. Although I don't think that's necessary. What I had in mind for the code in upstream.go was that when in single-server mode, if an OP force-nicks you to something else (a few servers let you do that), then you get that information and your downstream nick is updated. It wasn't about waiting for the server ACK.

What do you think?

delthas:
👍 wrt the fixup commit, agreed that using `*m` everywhere can get confusing.

delthas:
Updated with 2 new commits, as discussed on IRC.

delthas:
Updated with more commits and some bug fixes. 😊

downstream.go
```diff
@@ -268,7 +268,7 @@ func (dc *downstreamConn) SendMessage(msg *irc.Message) {
 
 // marshalMessage re-formats a message coming from an upstream connection so
 // that it's suitable for being sent on this downstream connection. Only
-// messages that may appear in logs are supported.
```
emersion, line 271: Hmm. So previously we had all the marshalling happening in one place, and now it's scattered around again. Can we move the marshalling logic in the upstream MODE handler here?

Oh, but then we need to parse the mode twice, which is annoying. Maybe we could have a function that turns a mode string into a `map[byte]bool` indicating the mode changes? Then we could use it from both places.

upstream.go
```diff
@@ -381,11 +381,6 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 			if uc.requestSASL() {
 				break // we'll send CAP END after authentication is completed
 			}
-
-			uc.SendMessage(&irc.Message{
```
emersion, line 385: I wonder if there's a point in waiting for CAP ACK/NAK if we don't use SASL.

upstream.go
```diff
@@ -1514,6 +1520,8 @@ func (uc *upstreamConn) SendMessageLabeled(downstreamID uint64, msg *irc.Message
 		}
 		msg.Tags["label"] = irc.TagValue(fmt.Sprintf("sd-%d-%d", downstreamID, uc.nextLabelID))
 		uc.nextLabelID++
+	} else if !uc.caps["message-tags"] {
+		msg.Tags = nil
```
emersion, line 1524: Can we define `func (uc *upstreamConn) SendMessage()` and do this juggling there?

upstream.go
```diff
@@ -341,18 +347,20 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
 				target = msg.Prefix.Name
 			}
 			uc.produce(target, msg, nil)
 
-			highlight := msg.Prefix.Name != uc.nick && isHighlight(text, uc.nick)
-			if ch, ok := uc.network.channels[target]; ok && ch.Detached && highlight {
-				uc.forEachDownstream(func(dc *downstreamConn) {
-					sendServiceNOTICE(dc, fmt.Sprintf("highlight in %v: <%v> %v", dc.marshalEntity(uc.network, ch.Name), msg.Prefix.Name, text))
-				})
+			if msg.Command != "TAGMSG" {
```
emersion, line 357: Nit: we shouldn't need this `if`

downstream.go
```diff
@@ -1448,10 +1454,33 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
...
+			return err
+		}
+		tags := copyClientTags(msg.Tags)
+
+		for _, name := range strings.Split(targetsStr, ",") {
+			uc, upstreamName, err := dc.unmarshalEntity(name)
+			if err != nil {
+				return err
+			}
+			if !uc.caps["message-tags"] {
```
emersion, line 1474: This should already be taken care of in `SendMessage`. But maybe we should only have the check in one place?

upstream.go
```diff
@@ -692,7 +692,8 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 				if err != nil {
 					return err
 				}
-				ch.Members[msg.Prefix.Name] = nil
+				var m memberships
+				ch.Members[msg.Prefix.Name] = &m
```
emersion, line 710: Nit: can be simplified as `&memberships{}`

downstream.go
delthas: To marshal the MODE message we need to know which mode letters are channel modes and which ones are membership prefixes. Then we marshal the mode params for a membership prefix but not for a channel mode.

The set of membership prefixes is stored in `upstreamConn.availableMemberships`, which is only available from an upstream, ie when we're connected (which makes sense: that's an `RPL_ISUPPORT` we parse from the upstream after registration and may vary between connections). That's incompatible with `downstreamConn.marshalMessage`, which expects a `network`, which makes sense because `marshalMessage` is also used to marshal messages from a network we're not currently connected to. In our case we'll only send the message when we're currently connected to that upstream.

Hence the need for specific logic here.

delthas:
Updated to address comments.

downstream.go
emersion: All right, fair enough. I wonder if we can come up with a better design overall.

emersion:
Thanks a lot for your patience. Pushed the first 3 commits (thanks for reordering!). The remaining commits should be more straightforward.

downstream.go
```diff
@@ -777,6 +777,15 @@ func (dc *downstreamConn) welcome() error {
 		Params:  []string{dc.nick, "No MOTD"},
 	})
 
+	if uc := dc.upstream(); uc != nil && uc.nick != dc.nick {
```
emersion, line 783: Maybe we should have an `updateNick` function that does this, similar to `updateSupportedCaps`?

downstream.go
emersion: We need to make sure not to double-send the nick change.

upstream.go
```diff
@@ -1278,6 +1259,35 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
+		if err := parseMessageParams(msg, nil, &command, &reason); err != nil {
+			return err
+		}
+
+		if command == "LIST" {
+			ok := uc.endPendingLISTs(false)
+			if !ok {
+				return fmt.Errorf("unexpected response for LIST: %q: no matching pending LIST", msg.Command)
+			}
+			uc.forEachDownstreamByID(downstreamID, func(dc *downstreamConn) {
```
emersion, line 1279: This is duplicated in the `else` branch, we can move this out of the `if` instead

upstream.go
```diff
@@ -1298,6 +1308,27 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
 		// Ignore
 	default:
 		uc.logger.Printf("unhandled message: %v", msg)
+		if downstreamID != 0 {
+			uc.forEachDownstreamByID(downstreamID, func(dc *downstreamConn) {
+				var params []string
+				if len(msg.Params) > 0 {
+					// best effort marshaling for unknown messages, replies and errors:
```
emersion, line 1319: Hmm. Check the message is a numeric instead?

emersion:
I originally pushed only some commits, but in the end all of the commits LGTM. All pushed to master, thanks!

PR #17, opened 2020-05-01
Fix joining only one saved channel per network

delthas:
This fixes a serious bug added in 276ce12e, where in newNetwork all
channels point to the same channel, which causes soju to only join a
single channel when connecting to an upstream network.

This also adds the same kind of reassignment of a for loop variable in
user.run(), even though that function currently works correctly, as a
sanity improvement in case this function is changed in the future.


delthas:
This is a pretty important bug, I suggest reviewing it before my other PR.

PR #19, opened 2020-05-28
Add support for downstream CHATHISTORY

delthas:


logger.go
```diff
@@ -132,3 +133,74 @@ func formatMessage(msg *irc.Message) string {
 		return ""
 	}
 }
+
+type historyMessage struct {
```
emersion, line 137: Maybe we could just use `irc.Message` here instead?

downstream.go
```diff
@@ -1510,6 +1524,116 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
...
+				Params:  []string{"+" + batchRef, "chathistory", target},
+			})
+
+			history := make([]historyMessage, limit)
+			remaining := limit
+
+			first := true
+			tries := 0
+			for remaining > 0 {
+				buf, err := parseMessages(uc.network, entity, timestamp, remaining, first)
```
emersion, line 1592: Why do we need the first call to behave differently?

logger.go
```diff
@@ -132,3 +133,74 @@ func formatMessage(msg *irc.Message) string {
...
+	time    time.Time
+	command string
+	sender  string
+	text    string
+}
+
+func parseMessages(network *network, entity string, timestamp time.Time, count int, untilTimestamp bool) ([]historyMessage, error) {
```
emersion, line 144: Nit: consider s/count/limit/

emersion, line 144: Nit: consider renaming this function to `parseMessagesBefore`

downstream.go
delthas: For the first call, we only want messages up to a certain hour, minute, second. For the next calls (the previous days), we want messages up to the end of that day. (In other words, if we're searching before Jan 2 15:04:05, then the call for Jan 2 will search until 15:04:05, whereas thel calls for Jan 1, Dec 31, ... will search until midnight.)

logger.go
delthas: Right. Using historyMessage is a tad more efficient because we skip formatting the time of that message, but even though this is a performance-important part of the code because we call this on a lot of messages, it's probably not worth the effort using that intermediate struct.

downstream.go
emersion: What about replacing this:

```go
timestamp = timestamp.AddDate(0, 0, -1)
```

with this:

```go
t = time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, t.Location()).Add(-1)
```

That way, no need for a special case.

delthas: This is simpler, but it would incorrectly skip over messages posted on the very last nanosecond of that day. Which solution should I pick?

emersion: I think this trade-off is reasonable. Simplicity, correctness, performance, in this order.

delthas: Makes sense. I'll implement the simple solution then.

PR #20, opened 2020-06-06
Multiple features

delthas:

### DIFF
--- a/cmd/sojuctl/main.go
+++ b/cmd/sojuctl/main.go
@@ -120,7 +120,12 @@ func readPassword() ([]byte, error) {
 	} else {
 		fmt.Fprintf(os.Stderr, "Warning: Reading password from stdin.\n")
 		scanner := bufio.NewScanner(os.Stdin)
-		scanner.Scan()
+		if !scanner.Scan() {
+			if err := scanner.Err(); err != nil {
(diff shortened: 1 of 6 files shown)


doc/soju.1.scd
```diff
@@ -154,6 +154,9 @@ abbreviated form, for instance *network* can be abbreviated as *net* or just
 *change-password* <new password>
 	Change current user password.
 
+*create-user* -username <username> -password <password> [admin]
```
emersion, line 157: Missing dash in `[admin]`: it should read `[-admin]`

cmd/sojuctl/main.go
```diff
@@ -15,7 +15,7 @@ import (
 
 const usage = `usage: sojuctl [-config path] <action> [options...]
 
-  create-user <username>	Create a new user
+  create-user <username> [-admin]	Create a new user
```
emersion, line 18: Nit: this messes up description alignment for other commands

doc/soju.1.scd
emersion, line 157: Ah, just thought that it may be better to rename this command to `user create`. This would be more consistent with `network create` and will be more natural once we have other user commands (`user delete`, `user update`, etc)

emersion:
"Add an admin flag to users" needs a rebase

PR #24, opened 2020-07-18
Implement IDLE

delthas:
See IDLE [1] for details.

[1]: https://todo.sr.ht/~emersion/soju/71

### DIFF
--- a/downstream.go
+++ b/downstream.go
@@ -193,6 +193,10 @@ type downstreamConn struct {
 	lastBatchRef uint64
 
 	saslServer sasl.Server
+
+	idle           bool
+	idleHighlights []string
+	idleMessages   []*irc.Message
 }
 
 func newDownstreamConn(srv *Server, ic ircConn, id uint64) *downstreamConn {
@@ -204,6 +208,7 @@ func newDownstreamConn(srv *Server, ic ircConn, id uint64) *downstreamConn {
 		id:            id,
 		supportedCaps: make(map[string]string),
 		caps:          make(map[string]bool),
+		idleMessages:  make([]*irc.Message, 0, srv.IdleLimit),
 	}
 	dc.hostname = remoteAddr
 	if host, _, err := net.SplitHostPort(dc.hostname); err == nil {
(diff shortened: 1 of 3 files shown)


downstream.go
```diff
@@ -256,6 +261,13 @@ func (dc *downstreamConn) readMessages(ch chan<- event) error {
 	return nil
 }
 
+func (dc *downstreamConn) flushIdleMessages() {
+	for _, m := range dc.idleMessages {
+		dc.conn.SendMessage(m)
+	}
+	dc.idleMessages = make([]*irc.Message, 0, dc.srv.IdleLimit)
```
emersion, line 268: Instead of re-allocating the slice, we can probably do something like this to set its length to zero:

```go
dc.idleMessages = dc.idleMessages[:0]
```

emersion:
Overall this looks good. I'd like to perform some real-world experiments to confirm this reduces battery usage.

hhirtz:
> It would be nice to add the extension spec to `doc/` too. And hide the functionality behind a vendored CAP, to avoid conflicts.

-1

This is already behind an ISUPPORT flag and doesn't need a CAP (reminder that the use case for caps are to allow the server to break the original protocol, which is not the case here).

hhirtz:
Rebased on latest master (stupid github won't accept `.patch` files as attachment):
[0001-Implement-IDLE-71.patch.txt](https://github.com/emersion/soju/files/6215458/0001-Implement-IDLE-71.patch.txt)

hhirtz:
Client impl (no relevant data to share though...........)

https://github.com/hhirtz/Discussions/blob/master/app/src/main/java/site/srht/taiite/discussions/IRCActivity.kt#L84

hhirtz:
Rebased: [0001-Implement-IDLE-71.patch.txt](https://github.com/emersion/soju/files/7264771/0001-Implement-IDLE-71.patch.txt)

emersion:
I think it would be good to split the part where highlights are set up into a separate ext. We'll need exactly the same for chat history and web push, it'd be a shame to duplicate the functionality.

hhirtz:
Also would be nice to add the spec in `doc/`

emersion:
Done

emersion:
I don't think this is a good way forward, I'd rather work on Web Push support, which should better integrate with mobile OSes.

PR #25, opened 2020-09-08
Add support for the extended-join capability

delthas:
This simple implementation only advertises extended-join to downstreams
when all upstreams support it.

In the future, it could be modified so that soju buffers incoming
upstream JOINs, sends a WHO, waits for the reply, and sends an extended
join to the downstream; so that soju could advertise that capability
even when some or all upstreams do not support it. This is not the case
in this commit.

### DIFF
--- a/downstream.go
+++ b/downstream.go
@@ -75,8 +75,9 @@ var permanentDownstreamCaps = map[string]string{
 // needAllDownstreamCaps is the list of downstream capabilities that
 // require support from all upstreams to be enabled
 var needAllDownstreamCaps = map[string]string{
-	"away-notify":  "",
-	"multi-prefix": "",
+	"away-notify":   "",
+	"extended-join": "",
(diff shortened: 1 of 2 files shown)


downstream.go
```diff
@@ -298,7 +299,12 @@ func (dc *downstreamConn) marshalMessage(msg *irc.Message, net *network) *irc.Me
 	case "NICK":
 		// Nick change for another user
 		msg.Params[0] = dc.marshalEntity(net, msg.Params[0])
-	case "JOIN", "PART":
+	case "JOIN":
+		msg.Params[0] = dc.marshalEntity(net, msg.Params[0])
+		if !dc.caps["extended-join"] {
```
emersion, line 304: I think this should be moved to `SendMessage`.

delthas: Updated

emersion (approved):
LGTM, thanks!

PR #26, opened 2020-11-08
Add support for auto-detaching, auto-reattaching

delthas:
~~- Fix clearing the channel key on /PART detach
- Introduce Channel.AutoDetach, which stores a number of seconds a
  channel is idle until it is automatically detached. Idle means no
  message and no reattach.
- Introduce Channel.DetachRelay, which stores a filter for which
  messages to relay through BouncerServ for detached channels.
- Introduce Channel.Reattach, which stores a filter for which messages
  to trigger a channel reattach on.
- Introduce MessageFilter, which matches either no message at all, only
  highlights, or any message.
- Introduce the service command `channel update`, that enables users to
  change any of these settings for their channels.

Would like to get ideas for better wordings of these new fields. Thoughts on the 10s per-user ticker for checking channels for `AutoDetach`?

I'll probably split those commits after we're doing discussing to avoid having to do many fixups as we're iterating. EDIT: man page as well when we're done discussing

I have successfully tested this locally.~~

### DIFF
--- a/db.go
+++ b/db.go
@@ -3,8 +3,10 @@ package soju
 import (
 	"database/sql"
 	"fmt"
+	"math"
 	"strings"
 	"sync"
+	"time"
 
 	_ "github.com/mattn/go-sqlite3"
 )
@@ -52,11 +54,40 @@ func (net *Network) GetName() string {
 	return net.Addr
 }
 
+type MessageFilter int
+
+const (
+	// TODO: use customizable user defaults for FilterDefault
+	FilterDefault MessageFilter = iota
+	FilterNone
+	FilterHighlight
+	FilterMessage
+)
+
+func parseFilter(filter string) (MessageFilter, error) {
+	switch filter {
+	case "default":
+		return FilterDefault, nil
+	case "none":
+		return FilterNone, nil
+	case "highlight":
+		return FilterHighlight, nil
+	case "message":
+		return FilterMessage, nil
+	}
+	return 0, fmt.Errorf("unknown filter: %q", filter)
+}
+
 type Channel struct {
 	ID       int64
 	Name     string
 	Key      string
 	Detached bool
+
+	RelayDetached MessageFilter
+	ReattachOn    MessageFilter
+	DetachAfter   time.Duration
+	DetachOn      MessageFilter
 }
 
 const schema = `
@@ -93,6 +124,10 @@ CREATE TABLE Channel (
 	name VARCHAR(255) NOT NULL,
 	key VARCHAR(255),
 	detached INTEGER NOT NULL DEFAULT 0,
+	relay_detached INTEGER NOT NULL DEFAULT 0,
+	reattach_on INTEGER NOT NULL DEFAULT 0,
+	detach_after INTEGER NOT NULL DEFAULT 0,
+	detach_on INTEGER NOT NULL DEFAULT 0,
 	FOREIGN KEY(network) REFERENCES Network(id),
 	UNIQUE(network, name)
 );
@@ -145,6 +180,12 @@ var migrations = []string{
 			JOIN User ON Network.user = User.username;
 		DROP TABLE Network;
 		ALTER TABLE NetworkNew RENAME TO Network;
+	`,
+	`
+		ALTER TABLE Channel ADD COLUMN relay_detached INTEGER NOT NULL DEFAULT 0;
+		ALTER TABLE Channel ADD COLUMN reattach_on INTEGER NOT NULL DEFAULT 0;
+		ALTER TABLE Channel ADD COLUMN detach_after INTEGER NOT NULL DEFAULT 0;
+		ALTER TABLE Channel ADD COLUMN detach_on INTEGER NOT NULL DEFAULT 0;
 	`,
 }
 
@@ -447,7 +488,7 @@ func (db *DB) ListChannels(networkID int64) ([]Channel, error) {
 	db.lock.RLock()
 	defer db.lock.RUnlock()
 
-	rows, err := db.db.Query(`SELECT id, name, key, detached
+	rows, err := db.db.Query(`SELECT id, name, key, detached, relay_detached, reattach_on, detach_after, detach_on
 		FROM Channel
 		WHERE network = ?`, networkID)
 	if err != nil {
@@ -459,10 +500,12 @@ func (db *DB) ListChannels(networkID int64) ([]Channel, error) {
 	for rows.Next() {
 		var ch Channel
 		var key sql.NullString
-		if err := rows.Scan(&ch.ID, &ch.Name, &key, &ch.Detached); err != nil {
+		var detachAfter int64
+		if err := rows.Scan(&ch.ID, &ch.Name, &key, &ch.Detached, &ch.RelayDetached, &ch.ReattachOn, &detachAfter, &ch.DetachOn); err != nil {
 			return nil, err
 		}
 		ch.Key = key.String
+		ch.DetachAfter = time.Duration(detachAfter) * time.Second
 		channels = append(channels, ch)
 	}
 	if err := rows.Err(); err != nil {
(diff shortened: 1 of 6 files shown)


db.go
```diff
@@ -52,11 +52,44 @@ func (net *Network) GetName() string {
 	return net.Addr
 }
 
+type MessageFilter int
+
+const (
+	FilterNone      MessageFilter = 0
+	FilterHighlight               = 1
```
emersion, line 59: If you use explicit values, you also need to specify explicitly the type. Otherwise the type of `FilterHighlight` and `FilterMessage` will be `int` instead of `MessageFilter`.

But here I think `iota` could be used instead.

db.go
```diff
@@ -52,11 +52,44 @@ func (net *Network) GetName() string {
 	return net.Addr
 }
 
+type MessageFilter int
+
+const (
+	FilterNone      MessageFilter = 0
```
emersion, line 58: I wonder if we should have a `FilterDefault` value to mean "user didn't specify anything". This could be used in the future to set a per-user preference instead of having to set the preference on each channel.

This would increase complexity a little bit.

user.go
```diff
@@ -348,124 +348,144 @@ func (u *user) run() {
...
+				}
+				detachTicker.Stop()
+				return
+			default:
+				u.srv.Logger.Printf("received unknown event type: %T", e)
 			}
-		case eventStop:
-			u.forEachDownstream(func(dc *downstreamConn) {
-				dc.Close()
+		case <-detachTicker.C:
```
emersion, line 475: Instead of a ticker, maybe we could have a `time.Timer` set to the next potential detach timeout?

user.go
```diff
@@ -348,124 +348,144 @@ func (u *user) run() {
...
-		switch e := e.(type) {
-		case eventUpstreamConnected:
-			uc := e.uc
+	detachTicker := time.NewTicker(10 * time.Second)
 
-			uc.network.conn = uc
+	for {
+		select {
+		case e := <-u.events:
+			switch e := e.(type) {
```
emersion, line 356: Maybe it's a good time to split this `switch` into a separate function…

db.go
```diff
@@ -52,11 +52,44 @@ func (net *Network) GetName() string {
...
+	return 0, fmt.Errorf("unknown filter: %q", filter)
+}
+
 type Channel struct {
 	ID       int64
 	Name     string
 	Key      string
 	Detached bool
+
+	DetachTimeout int64 // seconds
```
emersion, line 81: Even if the DB stores this as an int64 in seconds, maybe it would be better to make this a `time.Duration` field to make it easier to manipulate elsewhere?

upstream.go
```diff
@@ -403,14 +404,31 @@ func (uc *upstreamConn) handleMessage(msg *irc.Message) error {
...
 			highlight := msg.Prefix.Name != uc.nick && isHighlight(text, uc.nick)
-			if ch, ok := uc.network.channels[target]; ok && ch.Detached && highlight {
-				uc.forEachDownstream(func(dc *downstreamConn) {
-					sendServiceNOTICE(dc, fmt.Sprintf("highlight in %v: <%v> %v", dc.marshalEntity(uc.network, ch.Name), msg.Prefix.Name, text))
-				})
+			if ch, ok := uc.network.channels[target]; ok && ch.Detached {
+				if ch.DetachRelay == FilterMessage || (ch.DetachRelay == FilterHighlight && highlight) {
+					uc.forEachDownstream(func(dc *downstreamConn) {
+						if highlight {
+							sendServiceNOTICE(dc, fmt.Sprintf("highlight in %v: <%v> %v", dc.marshalEntity(uc.network, ch.Name), msg.Prefix.Name, text))
```
emersion, line 417: Pretty long indent level. Maybe it's time to extract this into a function.

service.go
```diff
@@ -218,6 +219,15 @@ func init() {
 			desc:   "change your password",
 			handle: handlePasswordChange,
 		},
+		"channel": {
+			children: serviceCommandSet{
+				"update": {
+					usage:  "<name> [-auto-detach duration] [-detach-relay none|highlight|message] [-reattach none|highlight|message]",
```
emersion, line 225: * In the future maybe we'll want to set a filter for auto-detach, e.g. auto-detach if no highlight for a while.
* Maybe `-auto-detach-after` would be a better name?
* Maybe `-reattach-on` would be a better name?

emersion (changes requested):
Overall looks pretty good!

user.go
delthas: We'd need to recompute the next potential detach timeout when a channel detach timeout changes, I'd rather not add this complexity

user.go
delthas: Not needed IMO

service.go
delthas: `-reattach-on` is nice. what about `-detach-after`?

delthas:
There's also a tricky issue with the way bouncer commands are parsed: shell-style comments are supported, meaning that:
```channel u #a/baze -relay-detached message```
is interpreted as 
```channel u #comment``` 
and fails with `error: expected at least one argument`

Quoting does work around the issue:
```channel u "#a/baze" -relay-detached message
updated channel "#a/baze"```
but is not ideal.

Should we tweak the parsing routine to not try to match comments?

File: service.go (hunk 7 above)
Comment (emersion) [reply service.go]: Works for me

File: user.go (hunk 3 above)
Comment (emersion) [reply user.go:475]: I'm not that comfortable with looping over the channels each 10s for each user.

What about this other plan: add a `*time.Timer` field to `upstreamConn`. Introduce a new `eventDetachChannel` event. If the channel is configured to auto-detach, create the timer like so:

```go
func (uc *upstreamChannel) setAutoDetach(dur time.Duration) {
    if uc.detachTimer != nil {
        uc.detachTimer.Stop()
        uc.detachTimer = nil
    }

    if dur == 0 {
        return
    }

    uc.detachTimer = time.AfterFunc(dur, func() {
        uc.net.user.events <- eventDetachChannel{…}
    })
}
```

This should be simpler than my previous proposal. Wdyt?

Side note, per https://github.com/golang/go/issues/28100, `Timer.C` is unused for timers created via `time.AfterFunc`.

emersion:
>Should we tweak the parsing routine to try not to match comments?

I agree we shouldn't try to recognize comments in here. Trying to escape `#` will prove difficult I think and cause more pain than anything (`#` should be quoted when not already escaped, except in single-quotes where nothing can be escaped, but single-quotes can be themselves escaped or appear in double-quotes, etc). I think we should just re-implement a simple quoted strings parser, it's not that much work.

Let's leave this for another PR though.

user.go
delthas: Hmmm. I like this better than looping every 10s. We'll have to call this function when joining, parting, receiving messages, and changing channel detach timeouts, but it's worth it

user.go
```diff
@@ -348,124 +364,140 @@ func (u *user) run() {
...
 
-	for e := range u.events {
-		switch e := e.(type) {
-		case eventUpstreamConnected:
-			uc := e.uc
+	detachTicker := time.NewTicker(10 * time.Second)
 
-			uc.network.conn = uc
+	for {
+		select {
```
emersion, line 370: We don't need this `select` anymore.

downstream.go
```diff
@@ -1517,6 +1537,12 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 				Params:  []string{upstreamName, text},
 			}
 			uc.produce(upstreamName, echoMsg, dc)
+
+			if uch, ok := uc.channels[upstreamName]; ok {
```
emersion, line 1541: Maybe we could have a `upstreamConn.updateChannelAutoDetach(name)` helper that does just that?

emersion:
>I'll probably split those commits after we're doing discussing to avoid having to do many fixups as we're iterating. EDIT: man page as well when we're done discussing

I'm pretty happy with the status quo, now would be a good time for this

service.go
```diff
@@ -218,6 +218,15 @@ func init() {
 			desc:   "change your password",
 			handle: handlePasswordChange,
 		},
+		"channel": {
+			children: serviceCommandSet{
+				"update": {
+					usage:  "<name> [-relay-detached <default|none|highlight|message>] [-reattach-on <default|none|highlight|message>] [-detach-after <duration>] [-detach-on <default|highlight|message>]",
```
emersion, line 224: Nit: missing `-detach-on none`

user.go
```diff
@@ -348,6 +362,8 @@ func (u *user) run() {
 		go network.run()
 	}
 
+	detachTicker := time.NewTicker(10 * time.Second)
```
emersion, line 365: This is now unused.

user.go
```diff
@@ -176,63 +181,72 @@ func (net *network) stop() {
...
+			Params:  []string{dc.marshalEntity(net, ch.Name), "Detach"},
+		})
+	})
+}
 
-			var uch *upstreamChannel
-			if net.conn != nil {
-				uch = net.conn.channels[ch.Name]
-			}
+func (net *network) attach(ch *Channel) {
```
emersion, line 208: Nice, good idea :)

emersion (changes requested):
Apart from these minor nits, this PR is good to go!

emersion (approved):
LGTM, thanks!

PR #27, opened 2021-01-29
Add support for the wip soju.im/read capability and READ command

delthas:
READ lets downstream clients share information between each other about
what messages have been read by other downstreams.

Each target/entity has an optional corresponding read receipt, which is
stored as a timestamp.

- When a downstream sends:
  READ #chan timestamp=2020-01-01T01:23:45.000Z
  the read receipt for that target is set to that date
- soju sends READ to downstreams:
  - on JOIN, if the client uses the soju.im/read capability
  - when the read receipt timestamp is set by any downstream

The read receipt date is clamped by the previous receipt date and the
current time.

### DIFF
--- a/db.go
+++ b/db.go
@@ -28,6 +28,9 @@ type Database interface {
 
 	ListDeliveryReceipts(ctx context.Context, networkID int64) ([]DeliveryReceipt, error)
 	StoreClientDeliveryReceipts(ctx context.Context, networkID int64, client string, receipts []DeliveryReceipt) error
+
+	GetReadReceipt(ctx context.Context, networkID int64, name string) (*ReadReceipt, error)
+	StoreReadReceipt(ctx context.Context, networkID int64, receipt *ReadReceipt) error
 }
 
 type MetricsCollectorDatabase interface {
@@ -179,4 +182,10 @@ type DeliveryReceipt struct {
 	Target        string // channel or nick
 	Client        string
 	InternalMsgID string
+}
+
+type ReadReceipt struct {
+	ID        int64
+	Target    string // channel or nick
+	Timestamp time.Time
 }
--- a/db_postgres.go
+++ b/db_postgres.go
@@ -77,6 +77,14 @@ CREATE TABLE "DeliveryReceipt" (
 	internal_msgid VARCHAR(255) NOT NULL,
 	UNIQUE(network, target, client)
 );
+
+CREATE TABLE "ReadReceipt" (
+	id SERIAL PRIMARY KEY,
+	network INTEGER NOT NULL REFERENCES "Network"(id) ON DELETE CASCADE,
+	target VARCHAR(255) NOT NULL,
+	timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
+	UNIQUE(network, target)
+);
 `
 
 var postgresMigrations = []string{
@@ -88,6 +96,15 @@ var postgresMigrations = []string{
 			ALTER COLUMN sasl_mechanism
 			TYPE sasl_mechanism
 			USING sasl_mechanism::sasl_mechanism;
+	`,
+	`
+		CREATE TABLE "ReadReceipt" (
+			id SERIAL PRIMARY KEY,
+			network INTEGER NOT NULL REFERENCES "Network"(id) ON DELETE CASCADE,
+			target VARCHAR(255) NOT NULL,
+			timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
+			UNIQUE(network, target)
+		);
 	`,
 }
 
@@ -499,4 +516,45 @@ func (db *PostgresDB) StoreClientDeliveryReceipts(ctx context.Context, networkID
 	}
 
 	return tx.Commit()
+}
+
+func (db *PostgresDB) GetReadReceipt(ctx context.Context, networkID int64, name string) (*ReadReceipt, error) {
+	ctx, cancel := context.WithTimeout(ctx, postgresQueryTimeout)
+	defer cancel()
+
+	receipt := &ReadReceipt{
+		Target: name,
+	}
+
+	row := db.db.QueryRowContext(ctx,
+		`SELECT id, timestamp FROM "ReadReceipt" WHERE network = $1 AND target = $2`,
+		networkID, name)
+	if err := row.Scan(&receipt.ID, &receipt.Timestamp); err != nil {
+		if err == sql.ErrNoRows {
+			return nil, nil
+		}
+		return nil, err
+	}
+	return receipt, nil
+}
+
+func (db *PostgresDB) StoreReadReceipt(ctx context.Context, networkID int64, receipt *ReadReceipt) error {
+	ctx, cancel := context.WithTimeout(ctx, postgresQueryTimeout)
+	defer cancel()
+
+	var err error
+	if receipt.ID != 0 {
+		_, err = db.db.ExecContext(ctx, `
+			UPDATE "ReadReceipt"
+			SET timestamp = $1
+			WHERE id = $2`,
+			receipt.Timestamp, receipt.ID)
+	} else {
+		err = db.db.QueryRowContext(ctx, `
+			INSERT INTO "ReadReceipt" (network, target, timestamp)
+			VALUES ($1, $2, $3)
+			RETURNING id`,
+			networkID, receipt.Target, receipt.Timestamp).Scan(&receipt.ID)
+	}
+	return err
 }
--- a/db_sqlite.go
+++ b/db_sqlite.go
@@ -70,6 +70,15 @@ CREATE TABLE DeliveryReceipt (
 	FOREIGN KEY(network) REFERENCES Network(id),
 	UNIQUE(network, target, client)
 );
+
+CREATE TABLE ReadReceipt (
+	id INTEGER PRIMARY KEY,
+	network INTEGER NOT NULL,
+	target TEXT NOT NULL,
+	timestamp TEXT NOT NULL,
+	FOREIGN KEY(network) REFERENCES Network(id),
+	UNIQUE(network, target)
+);
 `
 
 var sqliteMigrations = []string{
@@ -169,6 +178,16 @@ var sqliteMigrations = []string{
 			FROM Network;
 		DROP TABLE Network;
 		ALTER TABLE NetworkNew RENAME TO Network;
+	`,
+	`
+		CREATE TABLE ReadReceipt (
+			id INTEGER PRIMARY KEY,
+			network INTEGER NOT NULL,
+			target TEXT NOT NULL,
+			timestamp TEXT NOT NULL,
+			FOREIGN KEY(network) REFERENCES Network(id),
+			UNIQUE(network, target)
+		);
 	`,
 }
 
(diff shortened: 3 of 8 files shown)


db.go
```diff
@@ -531,7 +562,7 @@ func (db *DB) StoreChannel(networkID int64, ch *Channel) error {
 	} else {
 		var res sql.Result
 		res, err = db.db.Exec(`INSERT INTO Channel(network, name, key, detached, relay_detached, reattach_on, detach_after, detach_on)
-			VALUES (?, ?, ?, ?, ?, ?, ?, ?)`,
+			VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)`,
```
emersion, line 565: Remnant from a previous approach?

db.go
```diff
@@ -187,6 +209,15 @@ var migrations = []string{
 		ALTER TABLE Channel ADD COLUMN detach_after INTEGER NOT NULL DEFAULT 0;
 		ALTER TABLE Channel ADD COLUMN detach_on INTEGER NOT NULL DEFAULT 0;
 	`,
+	`
+		CREATE TABLE Receipt (
+			id INTEGER PRIMARY KEY,
+			network INTEGER NOT NULL,
+			name VARCHAR(255) NOT NULL,
+			type INTEGER NOT NULL,
+			timestamp INTEGER NOT NULL,
```
emersion, line 218: IIRC SQLite chokes on this final comma

emersion:
It would be nice to have a short document in `docs/` describing the new extension.

db.go
```diff
@@ -548,3 +579,44 @@ func (db *DB) DeleteChannel(id int64) error {
...
+
+	row := db.db.QueryRow("SELECT id, timestamp FROM Receipt WHERE network = ? AND type = ? AND name = ?", networkID, ReceiptRead, name)
+	var timestamp int64
+	if err := row.Scan(&receipt.ID, &timestamp); err != nil {
+		if err == sql.ErrNoRows {
+			return nil, nil
+		}
+		return nil, err
+	}
+	receipt.Timestamp = time.Unix(timestamp/1000000000, timestamp%1000000000)
```
emersion, line 600: The docs say:

> It is valid to pass nsec outside the range [0, 999999999].

So I think we can just do `time.Unix(0, timestamp)` here.

emersion:
Looks pretty solid.

Ideally we should have a client implementation to make sure we haven't missed anything. For example…

I wonder if we should guarantee that the server sends `READ` after any `JOIN`? Something like `<prefix> READ <target> *` if the server doesn't have any data. This could make it easier to clients to wait for the `READ` message to be received. I'm not 100% clear how the client logic will look like, but combined with `CHATHISTORY` clients may want to request history after the `READ` timestamp if any. Not knowing whether `READ` will be sent or not would make client logic more involved in this case.

db.go
```diff
@@ -131,6 +145,14 @@ CREATE TABLE Channel (
 	FOREIGN KEY(network) REFERENCES Network(id),
 	UNIQUE(network, name)
 );
+
+CREATE TABLE Receipt (
+	id INTEGER PRIMARY KEY,
+	network INTEGER NOT NULL,
+	name VARCHAR(255) NOT NULL,
+	type INTEGER NOT NULL,
+	timestamp INTEGER NOT NULL
```
emersion, line 154: Can add these constraints:

```sql
	FOREIGN KEY(network) REFERENCES Network(id),
	UNIQUE(network, target)
```

db.go
```diff
@@ -131,6 +145,14 @@ CREATE TABLE Channel (
 	FOREIGN KEY(network) REFERENCES Network(id),
 	UNIQUE(network, name)
 );
+
+CREATE TABLE Receipt (
+	id INTEGER PRIMARY KEY,
+	network INTEGER NOT NULL,
+	name VARCHAR(255) NOT NULL,
```
emersion, line 152: Maybe rename to "target" for consistency with the CHATHISTORY spec.

doc/read.md
```diff
@@ -0,0 +1,101 @@
...
+
+The server MUST reply to a successful `READ` command using a `READ` server command, or using an error message.
+
+#### `READ` server command
+
+When sent from a client, the `READ` command signals to the client that the last message read by the user, to the best knowledge of the server, has the specified timestamp. In that case, the command has the following syntax:
+
+    <prefix> READ <target> <timestamp>
+
+The `prefix` is the current server prefix.
```
emersion, line 42: Maybe it would make more sense to make `prefix` the current user's instead.

doc/read.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Command flows
+
+The server SHOULD send a `READ` command to a client in the following cases:
+- if the `soju.im/read` capability is negotiated, after it sends a server `JOIN` command to the client for a corresponding channel
+- if the `soju.im/read` capability is negotiated, after the last read timestamp of the target changes
+
+If the last read timestamp for a target is unknown or not yet set, it SHOULD not be sent.
```
hhirtz, line 50:
```suggestion
If the last read timestamp for a target is unknown or not yet set, it SHOULD NOT be sent.
```

doc/read.md
```diff
@@ -0,0 +1,101 @@
...
+~~~~
+[s] :nick!ident@host JOIN #channel
+[s] :irc.host READ #channel timestamp=2019-01-04T14:33:26.123Z
+~~~~
+
+## Use Cases
+
+Clients can know whether a user has already read newly received messages. For clients that display notifications about new messages or highlights, knowing when messages have been read can enable them to clear notifications for messages that were already read on another device.
+
+Clients never have to actively get the read timestamp because it is provided to them on join and as updated by the server.
```
hhirtz, line 91: To be merged with `## Description`, or make a `## Motivation` section at the beginning of the spec.

doc/read.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Implementation Considerations
+
+Server implementations can typically store a per-target timestamp variable that stores the timestamp of the last read message. When it receives a new timestamp, it can clamp it between the last read timestamp and the current time, and broadcast the new value to all clients if it was changed.
+
+Client implementations can know when a user has read messages by using various techniques such as when the focus shifts to their window or activity, when the messages are scrolled, when the user is idle, etc. They should not assume that any message appended to the buffer is being read by the client right now, especially when the window does not have the focus or is not visible. It is indeed a best effort value.
+
+## Security Considerations
+
+No last read timestamp if ever exchanged with the actual upstream server the bouncer is connected to, so there is no privacy risk that the server might leak or use this read data to infer when the user is online.
```
hhirtz, line 101:
```suggestion
No last read timestamp is ever exchanged with the actual upstream server the bouncer is connected to, so there is no privacy risk that the server might leak or use this read data to infer when the user is online.
```

hhirtz (changes requested):
I require changes.

doc/read.md
```diff
@@ -0,0 +1,101 @@
...
+
+    READ <target> <timestamp>
+
+The `target` parameter specifies a single buffer (channel or nickname).
+
+The `timestamp` parameter MUST have the format `timestamp=YYYY-MM-DDThh:mm:ss.sssZ`, as in the [server-time](https://github.com/ircv3/ircv3-specifications/blob/master/extensions/server-time-3.2.md) extension.
+
+#### `READ` client command
+
+When sent from a client, the `READ` command signals to the server that the last message read by the user, to the best knowledge of the client, has the specified timestamp. The timestamp MUST correspond to a previous message `time` tag.
```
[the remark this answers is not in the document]
delthas: I'm not sure this is needed here. Maybe in one of the non normative sections. I'd rather keep to strict description of the specification in those paragraphs

delthas:
Rebased. What should we do for non-channel targets, ie user queries? It's harder to send the READ tiemstamp automatically because we can't really know when to send them/when the downstream opens a bufifer for it. Maybe add an explciit READ client command to request a timestamp?

delthas:
senpai impl available at: https://git.sr.ht/~delthas/senpai/commit/b8ff96cdf13795e112f1b6130fa65c5933a05655

db_postgres.go
```diff
@@ -89,6 +89,15 @@ var postgresMigrations = []string{
 			TYPE sasl_mechanism
 			USING sasl_mechanism::sasl_mechanism;
 	`,
+	`
+		CREATE TABLE "ReadReceipt" (
+			id SERIAL PRIMARY KEY,
+			network INTEGER NOT NULL REFERENCES "Network"(id) ON DELETE CASCADE,
+			target VARCHAR(255) NOT NULL,
+			timestamp INTEGER NOT NULL,
```
emersion, line 97: `TIMESTAMP WITH TIME ZONE` should be used instead (and is compatible with `time.Time`).

db_sqlite.go
```diff
@@ -70,6 +70,15 @@ CREATE TABLE DeliveryReceipt (
 	FOREIGN KEY(network) REFERENCES Network(id),
 	UNIQUE(network, target, client)
 );
+
+CREATE TABLE ReadReceipt (
+	id INTEGER PRIMARY KEY,
+	network INTEGER NOT NULL,
+	target VARCHAR(255) NOT NULL,
+	timestamp INTEGER NOT NULL,
```
emersion, line 78: I'd prefer to pick one of the [standard ways](https://www.sqlite.org/datatype3.html#date_and_time_datatype) to store timestamps in SQLite, so that built-in functions and third-party tools can easily parse the field. For the precision we require, `TEXT` [should be okay](https://pkg.go.dev/github.com/mattn/go-sqlite3#SQLiteTimestampFormats) (and we should be able to pass plain `time.Time` values to the driver).

doc/read.md
```diff
@@ -0,0 +1,128 @@
...
+
+If the server receives a `READ` command with missing parameters, the `NEED_MORE_PARAMS` error code MUST be returned.
+
+    FAIL READ NEED_MORE_PARAMS :Missing parameters
+
+If the selectors or limit supplied were invalid, the `INVALID_PARAMS` error code SHOULD be returned.
+
+    FAIL READ INVALID_PARAMS [invalid_parameters] :Invalid parameters
+
+If the read timestamp cannot be set or returned due to an error, the `MESSAGE_ERROR` error code SHOULD be returned.
```
emersion, line 81: Nit: I think chathistory has been using the wording `MESSAGE_ERROR` as a way to indicate that the message store hit an error. Here no message store is involved, so maybe should rename this to something like `TEMPORARILY_UNAVAILABLE` (like the account-registration spec), `READ_ERROR` or `INTERNAL_ERROR`.

doc/read.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Errors and Warnings
+
+Errors are returned using the standard replies syntax.
+
+If the server receives a `READ` command with missing parameters, the `NEED_MORE_PARAMS` error code MUST be returned.
+
+    FAIL READ NEED_MORE_PARAMS :Missing parameters
+
+If the selectors or limit supplied were invalid, the `INVALID_PARAMS` error code SHOULD be returned.
```
emersion, line 77: There is no limit involved in this ext.

emersion (commented):
Looks very good to me! A last round of comments below.

emersion (approved):
LGTM.

gamja implementation done and merged!

Thanks a lot!

PR #28, opened 2021-05-24
Add support for SA* commands from InspIRCd

delthas:
These InspIRCd commands are documented at [1].

[1]: https://docs.inspircd.org/3/modules/


emersion:
Maybe related to: https://todo.sr.ht/~emersion/soju/120

It would be nice if we could just pass-through unknown commands (just like we pass through unknown messages we receive). Won't work well in multi-upstream mode ofc.

emersion:
0f3dd2f2b17d forwards unknown messages to upstreams in single-upstream mode.

emersion:
The passthrough logic is enough for nonstandard commands IMHO.

PR #31, opened 2021-06-29
Sort `channel status` channels by name

jaknel:
Ignores #'s while sorting

### DIFF
--- a/service.go
+++ b/service.go
@@ -818,9 +818,17 @@ func handleServiceChannelStatus(dc *downstreamConn, params []string) error {
 	n := 0
 
 	sendNetwork := func(net *network) {
+		var channels []*Channel
 		for _, entry := range net.channels.innerMap {
-			ch := entry.value.(*Channel)
+			channels = append(channels, entry.value.(*Channel))
+		}
 
+		sort.Slice(channels, func(i, j int) bool {
+			return strings.ReplaceAll(channels[i].Name, "#", "") <
+				strings.ReplaceAll(channels[j].Name, "#", "")
+		})
+
+		for _, ch := range channels {
 			var uch *upstreamChannel
 			if net.conn != nil {
 				uch = net.conn.channels.Value(ch.Name)


service.go
```diff
@@ -818,9 +818,17 @@ func handleServiceChannelStatus(dc *downstreamConn, params []string) error {
 	n := 0
 
 	sendNetwork := func(net *network) {
+		var channels []*Channel
 		for _, entry := range net.channels.innerMap {
-			ch := entry.value.(*Channel)
+			channels = append(channels, entry.value.(*Channel))
+		}
+
+		sort.SliceStable(channels, func(i, j int) bool {
```
emersion, line 826: Any reason why a stable sort is used here? Can't we just use `sort.Slice`?

service.go
```diff
@@ -818,9 +818,17 @@ func handleServiceChannelStatus(dc *downstreamConn, params []string) error {
...
+			return strings.ReplaceAll(channels[i].Name, "#", "") <
```
emersion, line 827: What is the purpose of trimming `#`? Shouldn't all channels begin with a `#` prefix, removing the need to trim it?

emersion (commented):
Thanks for your PR! Here are a few comments.

service.go
jaknel: Some servers, like on Libera.chat, use ## in some channel names.  I felt #abc and ##abc should be adjacent after sorting.

service.go
jaknel: sort.Slice is the correct choice. I used sort.SliceStable thinking it would prevent two channels on different networks from swapping positions except 1) networks are processed individually and 2) the slice is created every time the channel status command is called, so SliceStable has no benefit.

service.go
emersion: Okay, that's fine by me.

emersion:
Can you squash these two commits into a single one? (See https://git-rebase.io for more information.)

jaknel:
Squashed.

emersion (approved):
Thanks!

PR #35, opened 2021-09-29
Add support for draft/event-playback

delthas:


hhirtz (changes requested):
*hasn't tested the log parsing code*

msgstore_fs.go
```diff
@@ -246,38 +246,133 @@ func formatMessage(msg *irc.Message) string {
...
+			} else if r := strings.TrimPrefix(rem, "sets mode: "); r != rem {
+				cmd = "MODE"
+				prefix = &irc.Prefix{
+					Name: nick,
+				}
+				params = append([]string{entity}, strings.Split(r, " ")...)
+				// TODO: support marshaling this message even when disconnected from upstream
+				//       -> hard: channel mode types may change over time.
+				//       -> disable sending MODE messages until this is fixed.
+				return nil, time.Time{}, nil
```
delthas, line 342: Thoughts @emersion ? (If this return is not added, we pass this to marshalMessage, which panics.)

delthas:
Log parsing code updated & tested. Should be ready for review & merge, except for the specific comment above.

msgstore_fs.go
emersion: I think it'd be fine to make `marshalMessage` a no-op in single upstream mode.

downstream.go
```diff
@@ -904,6 +910,7 @@ func (dc *downstreamConn) updateSupportedCaps() {
 			supportedCaps[cap] = supported && uc.caps[cap]
 		}
 	})
+	supportedCaps["draft/event-playback"] = dc.srv.LogPath != "" && dc.network != nil
```
emersion, line 913: The `supportedCaps` variable is part of the `needAllDownstreamCaps` logic. I'd prefer to manually call `dc.setSupportedCap()` instead of abusing the `needAllDownstreamCaps` logic.

downstream.go
```diff
@@ -514,6 +516,10 @@ func (dc *downstreamConn) marshalMessage(msg *irc.Message, net *network) *irc.Me
 		msg.Params[0] = dc.marshalEntity(net, msg.Params[0])
 	case "QUIT", "SETNAME":
 		// This space is intentionally left blank
+	case "MODE":
+		if dc.network == nil {
+			return nil
+		}
```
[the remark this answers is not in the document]
emersion: Hm, can we just add this at the top of the function:

```go
if dc.network == nil {
    return msg
}
```

and then remove this `case`?

downstream.go
```diff
@@ -1321,10 +1331,6 @@ func (dc *downstreamConn) sendTargetBacklog(net *network, target, msgID string)
 
 	dc.SendBatch("chathistory", []string{dc.marshalEntity(net, target)}, nil, func(batchRef irc.TagValue) {
 		for _, msg := range history {
-			if !dc.messageSupportsHistory(msg) {
-				continue
-			}
-
```
hhirtz, line 1327: Why is this `if` block removed.

IIUC this will break clients that do not support CHATHISTORY when the backlog contains anything other than PRIVMSG and NOTICE messages.

hhirtz (approved):
I am not willing to test changes made to `msgstore_fs.go`. Otherwise, looks good to me.

Just one question, for science:

downstream.go
emersion: Hm that seems correct...

delthas: LoadLatestID only returns PRIVMSG & NOTICE messages by contract, so this check isnt needed

hhirtz: I see, that makes sense.

Then `memoryMessageStore`'s `Append` should be modified to drop non-event messages (I think they are useless in this message store? otherwise they should be dropped at the `LoadLatestID` level)

https://github.com/emersion/soju/blob/832d8b89a2e6a13bf850ca7641a7203cbe64b413/msgstore_memory.go#L76

emersion (approved):
LGTM, thanks!

PR #36, opened 2021-10-05
Add ircs+noverify to skip TLS verification

idanoo:


emersion:
Sorry, but I don't think this is desirable. Use `irc+insecure`, or implement TLS certificate pinning.

idanoo:
Thanks @emersion, will give Certpinning a go instead

PR #37, opened 2021-11-15
feat(CHATHISTORY): add LATEST support

vigoux:
My attempt at adding LATEST support for the CHATHISTORY command.

I am a beginner Go writter, and not very used to the codebase.

Also, I was not able to test this (which is bad I know), but I am going to.

### DIFF
--- a/downstream.go
+++ b/downstream.go
@@ -2261,7 +2261,7 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 		var target, limitStr string
 		var boundsStr [2]string
 		switch subcommand {
-		case "AFTER", "BEFORE":
+		case "AFTER", "BEFORE", "LATEST":
 			if err := parseMessageParams(msg, nil, &target, &boundsStr[0], &limitStr); err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


vigoux:
I tested it a tiny bit, and it seemed to work... I am not sure though about the "workload" I put on it, so I guess if someone else could test it, that would be great !

downstream.go
```diff
@@ -2304,7 +2304,7 @@ func (dc *downstreamConn) handleMessageRegistered(msg *irc.Message) error {
 		// TODO: support msgid criteria
 		var bounds [2]time.Time
 		bounds[0] = parseChatHistoryBound(boundsStr[0])
-		if bounds[0].IsZero() {
+		if bounds[0].IsZero() && !(subcommand == "LATEST" && boundsStr[0] == "*") {
```
emersion, line 2307: Maybe instead of having this special case, we can fill `bounds[0]`? Something among these lines:

```go
if subcommand == "LATEST" && boundsStr[0] == "*" {
    bounds[0] = time.Now()
}
```

With that, all of the logic below can stay as-is.

emersion:
Nitpick: we don't use the "feat(XXX)" notation for commit messages.

emersion:
Superseded by https://lists.sr.ht/~emersion/soju-dev/patches/26665

PR #38, opened 2022-02-15
Introduce draft+compression support

delthas:
Opening as draft for now, will need stats of real-life usage, and a spec writeup.


emersion:
For reference, some numbers were posted here: https://github.com/ircv3/ircv3-ideas/issues/66#issuecomment-1043018423

PR #41, opened 2022-07-25
wip: add soju.im/bouncer-buffers

delthas:


emersion:
I don't really think tying this to bouncer-networks is a good idea. I'd prefer to make the two completely separate (and allow e.g. Ergo to implement on but not the other).

I think we really need to figure out how to handle PMs before merging this.

delthas:
Will be done with #64 instead.

PR #45, opened 2022-12-27
Build default ISUPPORT passthrough values according to our defaults

delthas:
Previously, we would consider CHANTYPES to be #&+!, but would not store that entry in the ISUPPORT map, then causing soju to not send that value in its ISUPPORT to the downstream if it was not overwritten by the upstream.

Instead, we should advertise in our ISUPPORT the same defaults as those we use for our own logic.


emersion:
I don't think we should mutate the ISUPPORT sent by upstream servers when forwarding it to downstream clients.

PR #46, opened 2023-01-17
service: add user run

delthas:


service.go
```diff
@@ -35,12 +35,20 @@ var servicePrefix = &irc.Prefix{
 	Host: serviceNick,
 }
 
+type serviceContext struct {
+	context.Context
+	nick    string // optional
+	network *network
```
emersion, line 41: Can we also add `// optional` here?

service.go
```diff
@@ -1047,6 +1056,43 @@ func handleUserDelete(ctx *serviceContext, params []string) error {
...
+	ev := eventUserRun{
+		params: params,
+		print:  printCh,
+	}
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case u.events <- ev:
+	}
+	for text := range printCh {
```
emersion, line 1089: There is a race here:

- We send `ev` to `u.events`
- The user goroutine stops (because of a crash or user deletion)
- We block on `printCh`

To at least avoid blocking forever the admin goroutine, we can use a `select` when reading from `printCh`, and exit on `ctx.Done()`.

service.go
```diff
@@ -37,8 +37,8 @@ var servicePrefix = &irc.Prefix{
 
 type serviceContext struct {
 	context.Context
-	nick    string // optional
-	network *network
+	nick    string   // optional
+	network *network // optional
```
emersion, line 41: Nit: this belongs to the previous commit. I can take care of the rebasing if desirable.

service.go
emersion: (Ideally please add a TODO describing this issue, so that we can fix it properly later, probably by flushing the `u.events` queue and `close(ev.print)` in a `defer`.)

emersion (approved):
LGTM, thanks!

PR #47, opened 2023-01-17
service: add user status 

delthas:
Please apply https://github.com/emersion/soju/pull/46 first


service.go
```diff
@@ -874,6 +879,37 @@ func handleServiceSASLReset(ctx *serviceContext, params []string) error {
...
+		if len(users) == cap(users) {
+			break
+		}
+		users = append(users, user)
+	}
+	ctx.user.srv.lock.Unlock()
+
+	for _, user := range users {
+		line := user.Username
+		if user.Admin {
```
emersion, line 898: Hm. This is racy: the user goroutine may mutate `*user.User`.

Can we switch the slice to `[]database.User` instead? That way, we'd copy the fields, and wouldn't be subject to race conditions.

service.go
```diff
@@ -874,6 +879,37 @@ func handleServiceSASLReset(ctx *serviceContext, params []string) error {
...
+			line += " (admin)"
+		}
+		networks, err := ctx.user.srv.db.ListNetworks(ctx, user.ID)
+		if err != nil {
+			ctx.user.srv.lock.Unlock()
```
emersion, line 903: Remnant from a previous approach?

emersion:
Code looks good. Two minor things:

- Can the man page be updated as well?
- Can we print something like "(XXX more users omitted)" when there are more than 50 users?

delthas:
>     * Can the man page be updated as well?
> 
>     * Can we print something like "(XXX more users omitted)" when there are more than 50 users?

Done.

emersion (approved):
LGTM, thanks!

PR #49, opened 2023-03-20
add an example systemd service file

gdamjan:
The service file will run `/usr/bin/soju --config /etc/soju/config`.

By default it'll run as a dynamically created `soju` user and group (DynamicUser=yes), and it will run in a constrained environment[1], only having write access to /var/lib/soju, /var/log/soju and /run/soju

If the admin creates a static `soju` user and group, those static uid/gid will be used, with the same service constraints. This can be useful to share the static `soju` group with other services, so they can access the soju logs and database even if the service is not running (or before it's running).

The procedure for the initial user and database scaffold still works normally, when run as `root`:
```
mkdir /var/lib/soju/
sojudb -config /etc/soju/config create-user myname -admin
```
NOTE: systemd will automatically change the ownership of files in /var/lib/soju/ when the service is started.

`sojuctl` will work for root, or with `sudo -u soju`, just normally.

References:
- [1] https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=

### DIFF
--- a/contrib/soju.service
+++ b/contrib/soju.service
@@ -1,0 +1,21 @@
+[Unit]
+Description=soju IRC bouncer service
+Documentation=https://soju.im/
+Documentation=man:soju(1) man:sojuctl(1)
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=simple
+User=soju
+Group=soju
+DynamicUser=yes
+StateDirectory=soju
+ConfigurationDirectory=soju
+RuntimeDirectory=soju
+ExecStart=/usr/bin/soju
+ExecReload=/bin/kill -HUP $MAINPID
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target


gdamjan:
ps.
the soju config file (``/etc/soju/config`) will need to be world-readable, at least for the first start of the service. When the service is running it can be `chown`-ed to the soju user, and made mode `0700`.

In the future, once the Linux kernel supports `idmap` on tmpfs, and systemd adds the support for it,  `ConfigurationDirectory=soju` should be all that's needed for the soju process to have access to that directory.

contrib/soju.service
```diff
@@ -0,0 +1,22 @@
...
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=simple
+User=soju
+Group=soju
+DynamicUser=yes
+StateDirectory=soju
+LogsDirectory=soju
```
emersion, line 14: I don't think we actually need this: soju does not write logs on its own.

contrib/soju.service
```diff
@@ -0,0 +1,22 @@
...
+ConfigurationDirectory=soju
+RuntimeDirectory=soju
+ExecStart=/usr/bin/soju --config /etc/soju/config
```
emersion, line 17: The `-config` was required in the past, but is no longer necessary.

gdamjan: ah right, it's the default now. should I remove it then?

gdamjan:
a typical (and minimal) config file would be
```
db sqlite3 /var/lib/soju/main.db
log fs /var/lib/soju/logs/
listen irc+insecure://127.0.0.1:6667
listen unix+admin://
```

contrib/soju.service
emersion: Yes please!

contrib/soju.service
gdamjan: I can remove that.
I was thinking if it makes sense for an user to configure `log fs /var/log/soju/` ?

then they would be able to do `systemctl clean --what=logs soju.service`

not sure how useful that would be?!

emersion: This would be pretty confusing, the IRC logs should not be mixed up with application logs...

gdamjan: removed that too, the "message store" should probably **not** be in /var/log

gdamjan:
> > a typical (and minimal) config file
> 
> This should match `config.in`, installed by the Makefile.

indeed, I haven't seen the rename of "log" to "message-store" back in 2022 :scream:

emersion (approved):
Thanks!

PR #50, opened 2023-08-01
Fix SEARCH queries with special characters on SQLite

delthas:
Previously, we would pass SQLite FTS5 MATCH queries as is from the SEARCH parameter, when searching for messages.

By default FTS5 queries follow a specific syntax, e.g. AND, OR, ... can be used to filter searches logically.

Passing special characters in these unescaped queries, like `.`, would be rejected by SQLite, failing the SEARCH request.

Here, in order to replicate PostgreSQL and FS behavior, we simply want to search for the string as is, and not give it glob or logic operators semantics.

To do that, we quote each token of the string, skipping over separators.

### DIFF
--- a/database/sqlite.go
+++ b/database/sqlite.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"strings"
 	"time"
+	"unicode"
 
 	"git.sr.ht/~emersion/soju/xirc"
 	"github.com/prometheus/client_golang/prometheus"
@@ -1362,7 +1363,7 @@ func (db *SqliteDB) ListMessages(ctx context.Context, networkID int64, name stri
 		sql.Named("after", sqliteTime{options.AfterTime}),
 		sql.Named("before", sqliteTime{options.BeforeTime}),
 		sql.Named("sender", options.Sender),
-		sql.Named("text", options.Text),
+		sql.Named("text", quoteFTSQuery(options.Text)),
 		sql.Named("limit", options.Limit),
 	)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


emersion:
Hm, I thought this was by design, to expose the full featureset of SQLite's search queries to the user?

The spec says:

> The actual algorithm used for matching the text is implementation defined.

delthas:
Could be, but right now if you search for `example.com` you get an INTERNAL_ERROR because it is an invalid FTS query, which could be very unintuitive

database/sqlite.go
```diff
@@ -1398,3 +1399,50 @@ func (db *SqliteDB) ListMessages(ctx context.Context, networkID int64, name stri
...
+	// [...]
+	// FTS5 features three built-in tokenizer modules [...]:
+	// * The unicode61 tokenizer, based on the Unicode 6.1 standard. This is the default.
+	// [...]
+	// The unicode tokenizer classifies all unicode characters as either "separator" or "token" characters. [...]
+	// All unicode characters assigned to a general category beginning with "L" or "N" (letters and numbers,
+	// specifically) or to category "Co" ("other, private use") are considered tokens.
+	// All other characters are separators.
+	var sb strings.Builder
+	tokens := ftsQuerySeparators.Split(query, -1)
```
emersion, line 1435: Instead of a regexp, I'd prefer to use `strings.FieldsFunc` together with `unicode.In(ch, unicode.L, unicode.N, unicode.Co)`.

delthas:
> BTW, do you know if PostgreSQL is doing something similar here, or are we good on that front?

Hehe, I went to check for that when doing the patch as well. For PostgreSQL we're OK because we do [`plainto_tsquery`](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES) so it parses the string as plain tokens IIUC. 

> plainto_tsquery transforms the unformatted text querytext to a tsquery value. The text is parsed and normalized much as for to_tsvector, then the & (AND) tsquery operator is inserted between surviving words.

database/sqlite.go
```diff
@@ -1398,3 +1399,49 @@ func (db *SqliteDB) ListMessages(ctx context.Context, networkID int64, name stri
...
+	// The unicode tokenizer classifies all unicode characters as either "separator" or "token" characters. [...]
+	// All unicode characters assigned to a general category beginning with "L" or "N" (letters and numbers,
+	// specifically) or to category "Co" ("other, private use") are considered tokens.
+	// All other characters are separators.
+	tokens := strings.FieldsFunc(query, func(r rune) bool {
+		return !unicode.In(r, unicode.L, unicode.N, unicode.Co)
+	})
+	var sb strings.Builder
+	for _, token := range tokens {
+		if token == "" {
```
emersion, line 1436: I think `strings.FieldsFunc` guarantees that this can't happen.

emersion (approved):
LGTM! Very nice commit message and explanation as usual. Thanks!

PR #52, opened 2023-08-13
Add support for solanum-specific caps

glguy:
Solanum implements a few useful caps for adding message tags to messages.

Specifications available at:
* https://solanum.chat/spec/identify-msg
* https://solanum.chat/spec/oper
* https://solanum.chat/spec/realhost


emersion:
I'm not in general opposed to the idea of merging support for vendored extensions, but I'd prefer to make an effort to standardize the extension before merging. IOW, only merge support for vendored exts as last resort when standardization fails.

glguy:
I think this approach simply makes soju less useful today. I'm willing to maintain a fork suitable for use with solanum in the meantime.

PR #53, opened 2023-08-13
Don't drop RPL_LISTSTART - passthrough to downstream

glguy:
Without this change, downstream clients never see the 321 reply that they'd expect before the list starts


emersion (approved):
Good catch, thanks!

Slightly edited to use a single `case`.

PR #55, opened 2023-08-23
Pass MODE and TOPIC through for unjoined channels

glguy:
I've tested that this passes this through, but there are code path's here I'm not at all familiar with yet, so please look at this very carefully if you weren't already doing that :)

### DIFF
--- a/upstream.go
+++ b/upstream.go
@@ -1282,14 +1282,15 @@ func (uc *upstreamConn) handleMessage(ctx context.Context, msg *irc.Message) err
 		if err := parseMessageParams(msg, nil, &name, &topic); err != nil {
 			return err
 		}
-		ch, err := uc.getChannel(name)
-		if err != nil {
-			return err
-		}
-		if msg.Command == irc.RPL_TOPIC {
-			ch.Topic = topic
+		ch := uc.channels.Get(name)
+		if ch == nil {
+			uc.forwardMsgByID(ctx, downstreamID, msg)
 		} else {
-			ch.Topic = ""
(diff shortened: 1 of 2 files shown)


upstream.go
```diff
@@ -1298,7 +1298,8 @@ func (uc *upstreamConn) handleMessage(ctx context.Context, msg *irc.Message) err
 		}
 		ch, err := uc.getChannel(name)
```
emersion, line 1299: Instead of `getChannel` and having to handle an error, we can do something like

```go
ch := uc.channels.Get(name)
if ch == nil {
    …
}
```

I'd slightly prefer that form because it's clearer when the code-path is taken. `err` could in theory be any runtime error.

emersion:
Ah, one last thing: we probably also want to pass-through `RPL_NOTOPIC` right?

upstream.go
```diff
@@ -1296,9 +1296,10 @@ func (uc *upstreamConn) handleMessage(ctx context.Context, msg *irc.Message) err
 		if err := parseMessageParams(msg, &name); err != nil {
 			return err
 		}
-		ch, err := uc.getChannel(name)
-		if err != nil {
-			return err
+		ch := uc.channels.Get(name)
+		if ch == nil {
+			uc.forwardMsgByID(ctx, downstreamID, msg)
+			return nil
```
emersion, line 1302: Hm, actually, this message is sent when the topic changes. When the client sends a `TOPIC` command to request the topic, I think `RPL_TOPIC` is used?

emersion (approved):
Thanks!

PR #57, opened 2023-08-29
Forward complete MODE message to upstream when changing modes

glguy:
On Solanum more arguments are meaningful when changing snomask flags: `MODE nick s +x`. I imagine there are other ircds that could do similar things


emersion (approved):
Thanks!

PR #59, opened 2023-09-29
Add support for 337 RPL_WHOISTEXT

glguy:
This message is in WHOIS replies for users with +I hide_idle_time set.

As seen on TV https://github.com/znc/znc/pull/1881


emersion:
Seems like this conflicts with RPL_ENDOFINVITELIST?

emersion:
Well, we'll just drop the RPL_ENDOFINVITELIST instead of forwarding it.

The server might send additional arbitrary WHOIS replies. Is it a big issue to broadcast these to downstream clients?

glguy:
> Well, we'll just drop the RPL_ENDOFINVITELIST instead of forwarding it.
> 
> The server might send additional arbitrary WHOIS replies. Is it a big issue to broadcast these to downstream clients?

It just depends on why the feature was implemented in the first place. If it's implemented to route replies to the client that requested the reply then this would be needed. If not a bunch of code could be removed

emersion:
Yes, it's implemented to route replies correctly, but it's not possible to do so for all non-standard server replies.

PR #61, opened 2024-04-30
Fix missing prefix on away numerics

jacob1:
Prefix seems to be required on numeric server responses. Without it, clients like hexchat can't parse it at all and ignore it. This prevents /back from working, hexchat won't let you unaway unless it parses the original /away

### DIFF
--- a/downstream.go
+++ b/downstream.go
@@ -752,7 +752,7 @@ func (dc *downstreamConn) handleMessageUnregistered(ctx context.Context, msg *ir
 			dc.away = nil
 		}
 
-		dc.SendMessage(ctx, generateAwayReply(dc.away != nil))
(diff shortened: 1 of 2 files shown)


emersion:
I believe this is a HexChat bug: https://modern.ircdocs.horse/#source

> Clients MUST be able to process any given message the same way whether it contains a source or does not contain one.

jacob1:
Yeah I was using the same site. I thought the same, until I read below under "Numeric Replies":

>Distinct from a normal message, a numeric reply MUST contain a &lt;source&gt; and use a three-digit numeric as the command

emersion:
Oh, good catch! Makes sense then, thanks!

PR #63, opened 2024-07-17
Update znc-import tool to tell the user to use sojudb not sojuctl to change passwords for imported users.

Prototyped:
change-password is a subcommand of sojudb. znc-import does not import user passwords and needs to tell the user to use that tool.


emersion (approved):
In general `sojuctl` is a better tool because it doesn't require the daemon to be stopped, however the import scripts does require the daemon to be stopped so `sojudb` is a better pick here indeed.

Thanks!

PR #64, opened 2024-08-27
Add support for channel pinned/muted status through METADATA

delthas:
Tested on my end with a senpai branch.

---

First, this implements support for the metadata-2 specification, work-in-progress by progval on ircv3-specifications [1].

The METADATA implementation is bouncer-only (namely, it does not interact with any upstream METADATA).

The implementation is specific in that there are only 2 supported keys: "pinned" and "muted". The user cannot set any other keys. Those keys can only be "0" or "1", representing a boolean.

This enables downstreams to get/set two flags on each "message target" (which can be either users or channels): pinned and muted.

- "pinned" is "1" when the channel should be pinned in the client, e.g. shown at the top of the channel list
- "muted" is "1" when the channel should be muted, e.g. have its notifications on non-highlights hidden

The actual specific client behavior for "pinned" and "muted" is not defined here. This just enabels clients to get/set the flags.

[1]: https://github.com/ircv3/ircv3-specifications/pull/501


database/postgres.go
```diff
@@ -10,11 +10,12 @@ import (
 	"strings"
 	"time"
 
-	"codeberg.org/emersion/soju/xirc"
 	_ "github.com/lib/pq"
 	"github.com/prometheus/client_golang/prometheus"
 	promcollectors "github.com/prometheus/client_golang/prometheus/collectors"
 	"gopkg.in/irc.v4"
+
+	"codeberg.org/emersion/soju/xirc"
```
delthas, line 18: Sorry for these noop changes. My formatter really wants to put the package import at the bottom. These can be removed before merging.

delthas:
Just for reference, this is an alternative to https://github.com/emersion/soju/pull/41

downstream.go
```diff
@@ -3322,6 +3563,41 @@ func (dc *downstreamConn) listWebPushSubscriptions(ctx context.Context) ([]datab
...
+		m["muted"] = "0"
+	}
+	dc.SendBatch(ctx, "metadata", nil, nil, func(batchRef string) {
+		for k := range dc.metadataSubs {
+			v, ok := m[k]
+			if !ok {
+				continue
+			}
+			dc.SendMessage(ctx, &irc.Message{
+				Prefix:  dc.prefix(),
```
emersion, line 3592: This can be left out now.

downstream.go
```diff
@@ -757,6 +763,54 @@ func (dc *downstreamConn) handleMessageUnregistered(ctx context.Context, msg *ir
 		}
 
 		dc.SendMessage(ctx, generateAwayReply(dc.away != nil))
+	case "METADATA":
+		var subcommand string
+		if err := parseMessageParams(msg, nil, &subcommand); err != nil {
+			return err
+		}
+		switch subcommand {
```
emersion, line 771: Seems like this `switch` is identical pre-conn-reg and post-conn-reg - could we dedup these?

downstream.go
```diff
@@ -2679,6 +2736,190 @@ func (dc *downstreamConn) handleMessageRegistered(ctx context.Context, msg *irc.
...
+				})
+			}
+			return nil
+		}
+		if dc.network == nil || target == "*" {
+			return ircError{&irc.Message{
+				Command: "FAIL",
+				Params:  []string{"METADATA", "INVALID_TARGET", target, "Invalid target"},
+			}}
+		}
```
emersion, line 2787: We should probably send an empty list of `GET`/`LIST` instead of an error.

downstream.go
```diff
@@ -2679,6 +2736,190 @@ func (dc *downstreamConn) handleMessageRegistered(ctx context.Context, msg *irc.
...
+			if mt.Pinned {
+				m["pinned"] = "1"
+			} else {
+				m["pinned"] = "0"
+			}
+			if mt.Muted {
+				m["muted"] = "1"
+			} else {
+				m["muted"] = "0"
+			}
```
emersion, line 2808: Could extract this in a helper function since it's shared with `sendChannelMetadata`.

downstream.go
```diff
@@ -2679,6 +2736,190 @@ func (dc *downstreamConn) handleMessageRegistered(ctx context.Context, msg *irc.
...
+				for k, v := range m {
+					var mv string
+					switch k {
+					case "pinned":
+						if v != nil && *v == "1" {
+							mt.Pinned = true
+							mv = "1"
+						} else {
+							mt.Pinned = false
+							mv = "0"
```
emersion, line 2871: Would prefer to error out if the value is neither `1` nor `0`.

downstream.go
```diff
@@ -2679,6 +2736,190 @@ func (dc *downstreamConn) handleMessageRegistered(ctx context.Context, msg *irc.
...
+						continue
+					}
+					dc.SendMessage(ctx, &irc.Message{
+						Tags:    tags,
+						Command: "761", // RPL_KEYVALUE
+						Params:  []string{dc.nick, target, k, "*", mv},
+					})
+					dc.network.forEachDownstream(func(dc *downstreamConn) {
+						dc.SendMessage(ctx, &irc.Message{
+							Prefix:  dc.prefix(),
```
emersion, line 2896: We shouldn't need to set the prefix here.

downstream.go
```diff
@@ -2679,6 +2736,190 @@ func (dc *downstreamConn) handleMessageRegistered(ctx context.Context, msg *irc.
...
+				var v *string
+				if len(msg.Params) > 3 {
+					v = &msg.Params[3]
+				}
+				m[k] = v
+			case "CLEAR":
+				m["pinned"] = nil
+				m["muted"] = nil
+			}
+			f := func(batchRef string) {
```
emersion, line 2857: Would be nicer to extract this to a helper method or something?

downstream.go
```diff
@@ -227,6 +227,7 @@ var permanentDownstreamCaps = map[string]string{
 	"server-time":   "",
 	"setname":       "",
 
+	"draft/metadata-2":        "before-connect,max-keys=0,max-value-bytes=1",
```
emersion, line 230: We're missing `draft/metadata-notify-2` here I think.

downstream.go
```diff
@@ -3398,6 +3674,8 @@ func forwardChannel(ctx context.Context, dc *downstreamConn, ch *upstreamChannel
 	if !dc.caps.IsEnabled("soju.im/no-implicit-names") && !dc.caps.IsEnabled("draft/no-implicit-names") {
 		sendNames(ctx, dc, ch)
 	}
+
+	dc.sendChannelMetadata(ctx, ch.Name)
```
emersion, line 3678: This handles channel metadata, but we're missing support for users. We need to check for monitored users and send updates accordingly if `draft/metadata-notify-2` is enabled.

downstream.go
delthas: Will be done in a separate MR:

> If the draft/metadata-notify-2 capability is requested, clients also receive metadata updates for the users they are currently monitoring.

We do not support that yet

downstream.go
delthas: Will be done in a separate MR

delthas:
Updated, ready for review.

delthas:
Gentle ping 🎀

downstream.go
```diff
@@ -2714,6 +2737,104 @@ func (dc *downstreamConn) handleMessageRegistered(ctx context.Context, msg *irc.
...
+		if err := parseMessageParams(msg, &target, &subcommand); err != nil {
+			return err
+		}
+		if handled, err := dc.handleMetadataSub(ctx, msg); handled {
+			return err
+		}
+		var mt *database.MessageTarget
+		if dc.network != nil && target != "*" {
+			var err error
+			mt, err = dc.srv.db.GetMessageTarget(ctx, dc.network.ID, target)
```
emersion, line 2751: Need to `network.casemap()` target names here (other places too probably)

delthas:
Gentle ping 🎀

emersion (approved):
Thanks a bunch!

PR #67, opened 2025-05-01
contrib/nginx: new document

gdamjan:
document running soju behind a nginx proxy for both http/websockets and the tcp/irc protocol.

### DIFF
--- a/contrib/nginx.md
+++ b/contrib/nginx.md
@@ -1,0 +1,114 @@
+# Setting up nginx as a proxy to soju
+
+You can configure nginx as a proxy to soju to handle both public http/tls and also irc/tls connections.
+This way we don't need to expose soju listening sockets publicly and relieve it from handling tls.
+One benefit of this approach would be that soju doesn't need to be hooked into certbot, nor does it need
+to be reloaded when certificates are updated.
+
+In this example, nginx will also handle serving static files for gamja, from the `/srv/gamja` directory.
+
+**Note:** different Linux distributions use a different configuration structure for nginx, so here we only
+provide the configuration snippets for the [`http`/`server`](https://nginx.org/en/docs/http/ngx_http_core_module.html#server)
+module, the [http proxy](https://nginx.org/en/docs/http/ngx_http_proxy_module.html) configuration and
+the [`stream`/`server`](https://nginx.org/en/docs/stream/ngx_stream_core_module.html) module.
+
+Please refer to your OS documentation to learn how to configure nginx, and where its config files are located.
+
+## http server: websocket proxying
+
+First, configure Soju to listen for websocket on the `/run/soju/http.sock` unix domain socket:
+```
+…
+listen http+unix:///run/soju/http.sock
+…
+```
+
+Then configure nginx to proxy all http traffic for the `/socket` prefix to that unix domain socket,
+by using the `proxy_pass` directives:
+
+```
+# /etc/nginx/sites-enabled/60-soju.conf
+server {
+  server_name chat.example.com;
+  listen 80;
+  listen [::]:80;
+
+  location /socket {
+    proxy_pass http://unix:/run/soju/http.sock:/socket;
+    proxy_read_timeout 600s;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "Upgrade";
+    proxy_set_header Forwarded ""; # prevent clients from spoofing this header, which takes precedence over X-Forwarded-*
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header Host $host;
+  }
+
+  location /uploads {
+    proxy_pass http://unix:/run/soju/http.sock:/uploads;
+    proxy_set_header Forwarded ""; # prevent clients from spoofing this header, which takes precedence over X-Forwarded-*
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header Host $host;
+  }
+
+  root /srv/gamja/;
+}
+```
+All the rest of http requests will be served from staic files in the `/src/gamja` directory
+(you can just copy the gamja `dist/` production files there).
+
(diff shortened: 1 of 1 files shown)


contrib/nginx.md
```diff
@@ -0,0 +1,106 @@
+# Setting up nginx as a proxy to soju
+
+You can configure nginx as a proxy to soju to handle both public http/tls and also irc/tls connections.
+This way we don't need to expose soju listening sockets publicly and relieve it from handling tls.
+One benefit of this approach would be that soju doesn't need to be hooked into certbot, nor does it need
+to be reloaded when certificates are updated. Potentially the nginx TLS implementation might be more robust
+but that hasn't been proven :D.
```
emersion, line 7: I'd leave that last sentence out. The Go TLS implementation is widely used.

contrib/nginx.md
```diff
@@ -0,0 +1,106 @@
...
+Now configure nginx to listen on the 1667 tcp port (an an example), and proxy to soju using the "PROXY protocol".
+The `ssl_certificate`, `ssl_certificate_key` and `ssl_dhparam` are copy/pasted from the http server config
+(as edited by certbot).
+```
+# /etc/nginx/streams-enabled/60-soju.conf
+server {
+  listen      1667 ssl;
+  listen      [::]:1667 ssl;
+  proxy_pass unix:/run/soju/irc.sock;
+  proxy_protocol on;
```
emersion, line 80: Note: soju currently ignores the PROXY protocol for Unix sockets. We have `accept-proxy-ips` in the config file but no equivalent for Unix sockets. We need an explicit opt-in on the soju side: if the reverse proxy doesn't enable the PROXY protocol and just passes through bytes, any external user can send a fake PROXY header.

Nothing to change here because it's best to enable this on the nginx side even if the soju side doesn't deal with it yet.

contrib/nginx.md
```diff
@@ -0,0 +1,106 @@
...
+  listen [::]:80;
+
+  location /socket {
+    proxy_pass http://unix:/run/soju/ws.sock:/socket;
+    proxy_read_timeout 600s;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "Upgrade";
+    proxy_set_header X-Forwarded-For $remote_addr;
+    proxy_set_header X-Forwarded-Proto $scheme;
```
emersion, line 44: Note: it'd be better to use the standardized `Forwarded` header for this, however it seems like setting it up is pretty cumbersome with nginx. So probably best to stick with good ol' `X-Forwarded-*`.

contrib/nginx.md
```diff
@@ -0,0 +1,106 @@
...
+```
+drwxr-xr-x 2 soju www-data 100 May  1 03:00 /run/soju/
+srwxrwxr-x 1 soju www-data 0 May  1 03:00 /run/soju/irc.sock
+srwxrwxr-x 1 soju www-data 0 May  1 03:00 /run/soju/ws.sock
+```
+A common way to establish that when running soju under debian/ubuntu is to use the `Group=` directive
+in the soju .service unit file (or a drop-in):
+```
+[Service]
+Group=www-data
```
emersion, line 105: Maybe `SupplementaryGroups` would be better suited, so that the primary group isn't erased?

Hm, I suppose this may not be enough? Would it work to add a drop-in for nginx to add the soju supplementary group?

contrib/nginx.md
```diff
@@ -0,0 +1,106 @@
...
+the [`stream`/`server`](https://nginx.org/en/docs/stream/ngx_stream_core_module.html) module.
+
+Please refer to your OS documentation to learn how to configure nginx, and where its config files are located.
+
+## http server: websocket proxying
+
+First, configure Soju to listen for websocket on the `/run/soju/ws.sock` unix domain socket:
+```
+…
+listen ws+unix:///run/soju/ws.sock
```
emersion, line 23: In general `http+unix://` should be preferred so that file upload works fine as well. (This may require additional `location` blocks in the nginx config?)

emersion (commented):
Thanks a lot! Here are a few comments, but overall this looks very good.

contrib/nginx.md
gdamjan: > Maybe `SupplementaryGroups` would be better suited, so that the primary group isn't erased?

it can work like that, but then `soju` would need to explicitly set the group of the unix socket files.

> Hm, I suppose this may not be enough? Would it work to add a drop-in for nginx to add the soju supplementary group?

The nginx service runs as root and explicitly switches user/group by itself. not sure it supports supplementary groups either :/

contrib/nginx.md
gdamjan: good point, I haven't implemented file uploads on my instance, and it didn't pass my mind. let me see what's needed.

gdamjan: just to make it clear, `listen http+unix:///run/soju/ws.sock` will handle both `/socket` AND `/uploads` right?

emersion: Yup!

contrib/nginx.md
```diff
@@ -0,0 +1,112 @@
...
+    proxy_read_timeout 600s;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "Upgrade";
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header Host $host;
+  }
+
+  location /upload {
```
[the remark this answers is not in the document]
gdamjan: typo :facepalm:

contrib/nginx.md
```diff
@@ -0,0 +1,114 @@
...
+  listen 80;
+  listen [::]:80;
+
+  location /socket {
+    proxy_pass http://unix:/run/soju/http.sock:/socket;
+    proxy_read_timeout 600s;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "Upgrade";
+    proxy_set_header Forwarded "";
```
gdamjan, line 42: something like 
```suggestion
    proxy_set_header Forwarded ""; # prevent clients from spoofing the header and overriding X-Forwarded-*
```
perhaps?

emersion: Applied, with a very small adjustment.

emersion (approved):
Thanks!

